PR #8, opened 2019-03-08
Condition end date step

BNewing:
[Link to ticket](https://www.pivotaltracker.com/n/projects/2311302/stories/164006813)

This adds in the condition end date step for the cautions path.

### DIFF
--- a/app/services/caution_decision_tree.rb
+++ b/app/services/caution_decision_tree.rb
@@ -1,4 +1,5 @@
 class CautionDecisionTree < BaseDecisionTree
+  # rubocop:disable Metrics/CyclomaticComplexity
   def destination
     return next_step if next_step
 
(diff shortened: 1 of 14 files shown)


config/locales/en.yml
```diff
@@ -123,6 +135,16 @@ en:
             content_html: If you don’t know the date you were cautioned, you can get this information for free by asking the police for a <a href='https://www.gov.uk/copy-of-police-records'>subject access request (SAR)</a>.
           date_legend: Enter the date
           date_hint: For example, 29 1 2018
+      conditional_end_date:
```
zheileman: Can we add the copy as in the prototype/design? I know it might eventually change but it makes it easier to review and to test locally.

app/services/caution_decision_tree.rb
```diff
@@ -1,17 +1,33 @@
 class CautionDecisionTree < BaseDecisionTree
+  # rubocop:disable Metrics/CyclomaticComplexity
   def destination
     return next_step if next_step
 
     case step_name
+    when :caution_type
```
zheileman: Although in terms of code it makes no difference, it is easier to understand the decision flow if we maintain the steps ordered. So in this case, this should go at the end.

app/services/caution_decision_tree.rb
```diff
@@ -1,17 +1,33 @@
...
+      home
     else
       raise InvalidStep, "Invalid step '#{as || step_params}'"
     end
   end
+  # rubocop:enable Metrics/CyclomaticComplexity
+
+  private
+
+  def caution_type_route
```
zheileman: This method can be renamed to `after_caution_type` (or similar) to better express it is when you answer the "caution type" question, that you enter here. Otherwise looks like it is the actual route of the "caution type" question, which is not.

And for now is ok, but next week when I'm in the office we can rework a bit the `disclosure_check&.caution_type&.include? 'conditional'` part, probably exposing a `#conditional?` method in the CautionType value object that respond true/false and not having this logic here 😉

PR #11, opened 2019-03-13
add exit page

tmckernan:
[Link to the story](https://www.pivotaltracker.com/n/projects/2311302/stories/164490634)

Add the the following:
- add exit page for conviction path
- add conviction Decision tree
- remove any factory bot from persisting

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,6 +7,14 @@ def edit_step(name)
 end
 # :nocov:
 
+# :nocov:
+def show_step(name)
+  resource name,
+           only:       [:show],
+           controller: name
+end
+# :nocov:
+
 Rails.application.routes.draw do
   root 'home#index'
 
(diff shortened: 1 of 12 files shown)


config/routes.rb
```diff
@@ -7,6 +7,13 @@ def edit_step(name)
 end
 # :nocov:
 
+
+def show_step(name)
```
zheileman: Wrap this method in the above `:nocov:` annotation, like we did with `edit_step`, as we are not testing directly this methods, and otherwise it reduces our test coverage.

config/locales/en.yml
```diff
@@ -151,3 +151,13 @@ en:
       caution_type:
         edit:
           page_title: Caution types
+    conviction:
```
zheileman: For exit / static or content pages, I've found out it is easier to maintain, produce and "read", if we just use translated views, instead of locales, in this case a `show.en.html.erb`, as the locales can grow a lot or be complicated with html markup sometimes. Also when the content is still WIP or likely to change.

But it is perfectly fine also this. It is just a suggestion.

PR #12, opened 2019-03-13
update start page copy

tmckernan:
[Link to story](https://www.pivotaltracker.com/n/projects/2311302/stories/163976122)

Update start page copy to the following based on [prototype](https://ex-offenders.glitch.me/start.html?)

Screen Shot 2019-03-13 at 11 08 45
Screen Shot 2019-03-13 at 11 09 03

### DIFF
--- a/app/views/home/index.en.html.erb
+++ b/app/views/home/index.en.html.erb
@@ -1,4 +1,4 @@
-<% title 'Check when your criminal record expires - GOV.UK' %>
+<% title '' %>
 
 <div class="govuk-width-container">
   <main class="govuk-main-wrapper">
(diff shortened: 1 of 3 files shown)


app/views/home/index.en.html.erb
```diff
@@ -1,23 +1,73 @@
-<% title '' %>
+<% title 'Check when your criminal record expires - GOV.UK' %>
```
zheileman: The suffix `- GOV.UK` is added automatically, no need to provide it here.

zheileman: Actually, this should be left blank (as before) and change the locale key `en.service.name` to whatever the service name is.

PR #13, opened 2019-03-14
add caution result page

tmckernan:
[Link to story](https://www.pivotaltracker.com/n/projects/2311302/stories/164006866)

Add the the following:

- caution result page
- CautionResultPresenter to present the questions and answers for simple caution journey
 
Result page in the [prototype](http://localhost:3000/steps/caution/result), i have updated it to use gov.uk design system

Screen Shot 2019-03-14 at 10 47 01

### DIFF
--- a/app/presenters/caution_result_presenter.rb
+++ b/app/presenters/caution_result_presenter.rb
@@ -1,0 +1,24 @@
+class CautionResultPresenter
+  attr_reader :disclosure_check
+  QuestionAnswerRow = Struct.new(:question, :value)
+
+  def initialize(disclosure_check)
+    @disclosure_check = disclosure_check
+  end
+
+  def summary
+    caution_questions.map do |item|
+      next if disclosure_check[item].nil?
+
+      QuestionAnswerRow.new(item, disclosure_check[item])
+    end
+  end
+
+  def expiry_date
+    ExpiryDateCalculator.new(disclosure_check: disclosure_check).expiry_date
+  end
+
+  def caution_questions
+    [:kind, :caution_date, :under_age, :caution_date].freeze
+  end
+end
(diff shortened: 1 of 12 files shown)


app/presenters/caution_result_presenter.rb
```diff
@@ -0,0 +1,20 @@
+class CautionResultPresenter
+  attr_reader :disclosure_check
+  QuestionAnswerRow = Struct.new(:question, :value)
+
+  def initialize(disclosure_check)
+    @disclosure_check = disclosure_check
+  end
+
+  def summay
```
zheileman: Typo

config/locales/en.yml
```diff
@@ -151,3 +151,27 @@ en:
       caution_type:
         edit:
           page_title: Caution types
+
+  caution_results_html:
+    kind:
+      question: Criminal record type
+      answers:
+        caution: Caution
+        conviction:
```
zheileman: Something missing here?

app/views/steps/caution/result/show.html.erb
```diff
@@ -0,0 +1,44 @@
...
+          <li>the current law on when cautions and convictions expire</li>
+        </ul>
+
+        <p class="govuk-body">
+          This table shows the information you provided, which was used to work out the expiry date. If you entered any incorrect information, the expiry date will be incorrect.
+        </p>
+
+        <dl class="govuk-summary-list">
+          <%= render partial: 'steps/caution/result/shared/row',
+                     collection: @presenter, as: :row %>
```
zheileman: Perhaps better to be explicit and call here `@presenter.summary` instead of doing it in the controller.

That way we can even, if you want to, refactor a bit the presenter to have 2 methods, one for the `#summary` and another one for the `#expiry_date`.

app/views/steps/caution/result/shared/_row.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="govuk-summary-list__row">
+  <dt class="govuk-summary-list__key">
+    <%=t "caution_results_html.#{row.question.to_s}.question" %>
```
zheileman: I think it is not necessary to do `#to_s` here, as the interpolation of a symbol will cast a string. Can you double check? (same thing a few lines below).

PR #17, opened 2019-03-18
update ruby version to ruby 2.6.2

tmckernan:
Update ruby version to 2.6.2 as i can't find  a docker image for ruby 2.6.1.

https://hub.docker.com/_/ruby
Screen Shot 2019-03-18 at 15 54 02


zheileman:
@tmckernan For the sake of simplicity and as we don't plan to deliver this to production any time soon, I would suggest to go with version 2.6.0 as it is ready to use on our MOJ templates. And later on, decide if we want to bump or give time for the new version to be in the templates repo.

PR #26, opened 2019-03-22
Add a mechanism to warn users before reseting a session

tmckernan:
[Link to story](https://www.pivotaltracker.com/n/projects/2311302/stories/164768529)

The following has been completed in this PR:
- Mechanism to warn a user before resetting the session
- Give user the choose to continue with the current session or start a new one

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -1,3 +1,18 @@
 class HomeController < ApplicationController
+  before_action :existing_disclosure_check_warning, :reset_disclosure_check_session
+
   def index; end
+
+  private
+
+  def in_progress_enough?
+    current_disclosure_check&.in_progress? &&
+      current_disclosure_check.navigation_stack.size > 1
+  end
+
+  def existing_disclosure_check_warning
+    return unless in_progress_enough? && !params.key?(:new)
+
+    redirect_to warning_reset_session_path
+  end
 end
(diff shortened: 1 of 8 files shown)


app/controllers/home_controller.rb
```diff
@@ -1,3 +1,18 @@
 class HomeController < ApplicationController
+  before_action :existing_application_warning, :reset_disclosure_check_session
```
zheileman: For consistency better to name this `existing_disclosure_check_warning`

config/locales/en.yml
```diff
@@ -165,6 +165,13 @@ en:
           inset_text_paragraph_one: Breaking the conditions of a caution usually leads to prosecution.
           inset_text_paragraph_two: If you were convicted of an offence after breaking the conditions of your caution, you’ll need to start again and select conviction instead of caution.
 
+  warning:
+    reset_session:
+      page_title: Application in progress
```
zheileman: Instead of `application` better to call it `check` or `disclosure check` or something similar.
There are a few places that need to be changed.

spec/controllers/home_controller_spec.rb
```diff
@@ -1,12 +1,98 @@
...
+        it 'resets the c100_application session data' do
+          expect(session).to receive(:delete).with(:disclosure_check_id).ordered
+          expect(session).to receive(:delete).with(:last_seen).ordered
+          expect(session).to receive(:delete) # any other deletes
+          get :index
+        end
+      end
+    end
+
+    context 'when an existing screening application exists' do
```
zheileman: We don't have checks in `screening` status so this scenario can be removed.

PR #27, opened 2019-03-22
Accessibility fixes

BNewing:
There are a few things that don't quite match up with the Design System at the moment. This PR is to make sure everything aligns and that the full benefits of the Design System are being pulled across.

Things to address:

- [x] Missing id on `<main>` elements
- [x] Links missing Design System class
- [x] ~Fix links in error summary panel (they need to link to the form element with the error in and not the error message that appears just above them)~

Each of this issues will be addressed in a separate commit, and the rationale behind them will be in the commit message details.

This is a work in progress, and this list will keep updating (until I remove this line)


app/views/about/index.en.html.erb
```diff
@@ -2,18 +2,20 @@
 
 <div class="grid-row">
```
zheileman: We can get rid of this file, it was a test. The real index page is in `views/home/index`

BNewing: Ok great :D I'll remove it!

BNewing:
'Fix links in error summary panel (they need to link to the form element with the error in and not the error message that appears just above them)' will be addressed in a separate PR (#32) by @tmckernan, as it turned out to be quite complex

zheileman:
Didn't realised the `en.yml` still contains the original structure before the splitting into smaller files.
We should refactor it and remove everything that has been moved to separate files otherwise there will be duplications.

BNewing:
@tmckernan and I had a chat about it, and for now I've moved across the necessary changes to the separate yml files, but I'm leaving it as that because Tony will be removing the en.yml file in a separate PR he's working on :)

zheileman:
@BNewing sure, no problem at all.

PR #29, opened 2019-03-25
Split locale file  into several smaller files

tmckernan:
Locale file can be come very large, quickly an it becomes a chore to update the locale file.   I have separated the main section into different files to make editing/adding keys easier.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,7 @@ group :development, :test do
   gem 'listen', '>= 3.0.5', '< 3.2'
   gem 'pry-byebug'
   gem 'pry-rails'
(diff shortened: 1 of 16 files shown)


zheileman:
I think this might be a bit too much 🤣. I believe there must be a balance between having a single locale file with hundreds of lines, and having dozens of little locale files with just a couple lines.

Perhaps we should split it into functional categories instead of by pages.
For example, a file for the form elements (labels, hints, etc. and perhaps even the errors, as they are coupled to the form fields) and another file for the steps. Maybe we can split the steps by `caution` and `conviction`.

The rest might remain in the top level general local file.

But we can re-evaluate this organisation as we grow the service.

PR #34, opened 2019-04-01
add new step known caution date

tmckernan:
[Link to ticket](https://www.pivotaltracker.com/n/projects/2311302/stories/164953853)

Add the following 
- a step between caution or conviction question  and enter caution date
- migration to add known_caution_date to disclosure-check table.
- update caution decision tree to bypass entering caution date if not know
- update caution result page to display caution date if caution date is known else this:

> Your caution expired on the day you were given it

- update feature tests

### DIFF
--- a/app/views/steps/caution/known_caution_date/edit.html.erb
+++ b/app/views/steps/caution/known_caution_date/edit.html.erb
@@ -1,0 +1,24 @@
+<% title t('.page_title') %>
+
+<div class="govuk-width-container">
+  <%= step_header %>
+
+  <main class="govuk-main-wrapper">
+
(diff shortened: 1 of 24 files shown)


app/views/steps/caution/known_caution_date/edit.html.erb
```diff
@@ -0,0 +1,24 @@
+<% title t('.page_title') %>
+
+<div class="govuk-width-container">
+  <%= step_header %>
+
+  <main class="govuk-main-wrapper">
```
BNewing: Could we add `id="main-content"` to this?

PR #35, opened 2019-04-01
Proof of concept: error focus

zheileman:
Possible alternative way to do the linking to a specific form input to gain the focus, from the error summary, in a very decoupled way.

Only gotcha so far I've found is we have a dynamic form object (`caution_type_form`) which shows different radios based on a previous question (`under_age`), so with this solution we can't know where to link in a static way.

There are possible ways around this, but perhaps better to settle with the default behaviour (same we have now, link to the fieldset, not a specific radio) while we work on introducing more form objects as it might well be this solution can't handle all scenarios.

### DIFF
--- a/config/locales/en/helpers.yml
+++ b/config/locales/en/helpers.yml
@@ -49,5 +49,25 @@ en:
           youth_simple_caution: Youth Simple caution
           youth_conditional_caution: Youth Conditional caution
 
-
-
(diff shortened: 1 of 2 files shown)


zheileman:
@BNewing @tmckernan just an idea!
Please note: I've not merged latest work on master on purpose as I didn't want to remove or revert any of the changes already done to the errors helper.

PR #36, opened 2019-04-01
Add Step: Do you know the date you were convicted?

tmckernan:
[Link to story](https://www.pivotaltracker.com/n/projects/2311302)

Add the first step on the conviction journey.

### DIFF
--- a/app/views/steps/conviction/known_conviction_date/edit.html.erb
+++ b/app/views/steps/conviction/known_conviction_date/edit.html.erb
@@ -12,13 +12,11 @@
         <!-- The header is part of the radios legend by default, but can be configured -->
 
         <%= step_form @form_object do |f| %>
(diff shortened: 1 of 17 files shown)


app/views/steps/conviction/known_conviction_date/edit.html.erb
```diff
@@ -0,0 +1,22 @@
+<% title t('.page_title') %>
+
+<div class="govuk-width-container">
+  <%= step_header %>
+
+  <main class="govuk-main-wrapper">
```
BNewing: Could we add `id="main-content"` onto the `<main>` tag please :)? This is so that the skip link will work

tmckernan: I have add the `id="main-content"` onto the `<main>` tag.  Also updated the template to add the id when we generate new edit pages.

PR #44, opened 2019-04-11
Add step: Which discharge were you given?

tmckernan:
[Link to story](https://www.pivotaltracker.com/n/projects/2311302/stories/165273607)

Add step:  Which discharge were you given?

### DIFF
--- a/spec/services/conviction_decision_tree_spec.rb
+++ b/spec/services/conviction_decision_tree_spec.rb
@@ -53,6 +53,11 @@ RSpec.describe ConvictionDecisionTree do
     it { is_expected.to have_destination(:custodial_sentence, :edit) }
   end
 
+  context 'when the step is `conviction_type` equal discharge' do
+    let(:step_params) { { conviction_type: 'discharge' } }
+    it { is_expected.to have_destination(:discharge, :edit) }
+  end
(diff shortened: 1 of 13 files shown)


spec/services/conviction_decision_tree_spec.rb
```diff
@@ -53,6 +53,11 @@
     it { is_expected.to have_destination(:custodial_sentence, :edit) }
   end
 
+  context 'when the step is `conviction_type` equal discharge' do
```
zheileman: Not urgent, but we can consider refactor these specs as they grow, to read more like:

```ruby
context 'when the step is `conviction_type`' do
  let(:step_params) { { conviction_type: type } }

  context 'and type is `discharge`' do
    let(:type) { 'discharge' }
    it { is_expected.to have_destination(:discharge, :edit) }
  end

  context 'and type is `community_sentence`' do
    let(:type) { 'community_sentence' }
    it { is_expected.to have_destination(:community_order, :edit) }
  end
end
```

tmckernan: 👍 I was thinking that myself, i am going to refactor this once i have completed all the conviction types.

PR #56, opened 2019-06-07
Unify the convictions types

zheileman:
A single value-object to unify all the different conviction types, with ability to get the parent or the children, and this way, have a tree of convictions that can be easily traversed.

The main objective of this refactor will be to get rid of the individual conviction DB fields existing at the moment (community_order, custodial_sentence, discharge, etc.) and associated form objects, controllers and hopefully even views (but probably we will need to use partials to customise some copy).

This is still WIP but will be pushing the changes as I go, up to a point where we decide if this is worth doing or not (because maybe, we find out this will not simplify but instead make the code harder to read and follow).


zheileman:
@tmckernan I think with this example we can see it will be possible and quite easy to do this refactor. Only `Community Sentence` is implemented here, but to implement the rest will be just a matter of adding them to the Value Object, and amending the locales.

After that, a huge cleanup can be done to remove all DB fields, controllers, etc. for the individual convictions.

Please checkout the branch, do some runs to see how it works and take a look at the individual commits, but I think we can merge this and take it from there. Please note I've not touched anything in regards with the calculator.

PR #65, opened 2019-06-11
Add Step: Was the length of conviction given in weeks or months?

tmckernan:
Basic step length of conviction step without any styling.  The styling will be in the next PR

### DIFF
--- a/app/controllers/steps/conviction/conviction_length_controller.rb
+++ b/app/controllers/steps/conviction/conviction_length_controller.rb
@@ -1,0 +1,17 @@
+module Steps
+  module Conviction
+    class ConvictionLengthController < Steps::ConvictionStepController
+      def edit
+        @form_object = ConvictionLengthForm.new(
+          disclosure_check: current_disclosure_check,
+          conviction_length_type: current_disclosure_check.conviction_length_type,
+          conviction_length: current_disclosure_check.conviction_length
+        )
+      end
+
+      def update
+        update_and_advance(ConvictionLengthForm)
+      end
+    end
+  end
+end
--- a/app/forms/steps/conviction/conviction_length_form.rb
+++ b/app/forms/steps/conviction/conviction_length_form.rb
@@ -1,0 +1,26 @@
+module Steps
+  module Conviction
+    class ConvictionLengthForm < BaseForm
+      attribute :conviction_length, Integer
+      attribute :conviction_length_type, String
+
+      def self.choices
+        ConvictionLengthType.string_values
+      end
+
+      validates_inclusion_of :conviction_length_type, in: choices
+      validates_numericality_of :conviction_length, greater_than: 0, only_integer: true
+
+      private
+
+      def persist!
+        raise DisclosureCheckNotFound unless disclosure_check
+
+        disclosure_check.update(
+          conviction_length_type: conviction_length_type,
+          conviction_length: conviction_length
+        )
+      end
+    end
+  end
+end
--- a/app/views/steps/conviction/conviction_length/edit.html.erb
+++ b/app/views/steps/conviction/conviction_length/edit.html.erb
@@ -1,0 +1,23 @@
+<% title t('.page_title') %>
+
+<div class="govuk-width-container">
+  <%= step_header %>
+
+  <main id="main-content" class="govuk-main-wrapper">
+
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <%= error_summary %>
+
+        <!-- The header is part of the radios legend by default, but can be configured -->
+        <%= step_form @form_object do |f| %>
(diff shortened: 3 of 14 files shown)


app/controllers/steps/conviction/conviction_length_controller.rb
```diff
@@ -0,0 +1,16 @@
+module Steps
+  module Conviction
+    class ConvictionLengthController < Steps::ConvictionStepController
+      def edit
+        @form_object = ConvictionLengthForm.new(
+          disclosure_check: current_disclosure_check,
+          conviction_length: current_disclosure_check.conviction_length
```
zheileman: `conviction_length_type` is missing. You can use build instead so it gets automatically the attributes from the form, refer to C100.

app/forms/steps/conviction/conviction_length_form.rb
```diff
@@ -0,0 +1,26 @@
...
+    class ConvictionLengthForm < BaseForm
+      attribute :conviction_length, Integer
+      attribute :conviction_length_type, String
+
+      def self.choices
+        ConvictionLengthType.string_values
+      end
+
+      validates_inclusion_of :conviction_length_type, in: choices
+      validates(:conviction_length, numericality: { greater_than_or_equal_to: 1 }, allow_blank: true)
```
zheileman: why allow_blank? it is mandatory.

app/views/steps/conviction/conviction_length/edit.html.erb
```diff
@@ -0,0 +1,23 @@
...
+
+  <main id="main-content" class="govuk-main-wrapper">
+
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <%= error_summary %>
+
+        <!-- The header is part of the radios legend by default, but can be configured -->
+        <%= step_form @form_object do |f| %>
+          <%= f.radio_button_fieldset :conviction_length_type, inline: false, choices: Steps::Conviction::ConvictionLengthForm.choices %>
```
zheileman: This radios should be inline, `inline: true`

config/locales/en/conviction.yml
```diff
@@ -30,3 +30,7 @@ en:
           page_title: conviction date
           heading:  When did the conviction end?
           lead_text: Date of conviction
+      conviction_length:
+        edit:
+          page_title: conviction length
+          heading:  Was the length of conviction given in weeks or months?
```
zheileman:
```suggestion
          heading: Was the length of conviction given in weeks or months?
```

zheileman: If we are using the fieldset then we don't need heading, otherwise, leave it, but better not to duplicate if one of the two locales is no needed.

config/locales/en/helpers.yml
```diff
@@ -82,6 +84,9 @@ en:
           disqualification: You were banned from driving for a certain period of time
           endorsement: You were found guilty of a motoring offence, which is now included on your driver record
           penalty_points: You were given number of penalty points for a driving offence
+        conviction_length:
+          weeks: weeks
```
zheileman: I imagine we don't need hints for these radios.
Then you can do as I did for `custodial_sentence`, empty strings to not show the hint.

db/migrate/20190610093304_add_conviction_length_to_disclosure_check.rb
```diff
@@ -0,0 +1,6 @@
+class AddConvictionLengthToDisclosureCheck < ActiveRecord::Migration[5.2]
+  def change
+    add_column :disclosure_checks, :conviction_length_type, :string
+    add_column :disclosure_checks, :conviction_length, :integer, default: 0
```
zheileman: Please do not use a default. Otherwise on loading the form it will show in the input. Being an annoyance for the user to have to delete and change, or worse, making it confusing for the user.

spec/forms/steps/conviction/conviction_length_form_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    context 'Validation' do
+      context 'when conviction_length_type is not given' do
+        let(:conviction_length_type) { nil }
+
+        it 'returns false' do
+          expect(subject.save).to be(false)
+        end
+
+        it 'has a validation error on the field' do
+          expect(subject).to_not be_valid
```
zheileman: This is not testing the appropriate validation error has been added to the field.
Please refer to for an example: https://github.com/ministryofjustice/disclosure-checker/blob/master/spec/forms/steps/conviction/known_date_form_spec.rb#L34

Same for validations below.

spec/services/conviction_decision_tree_spec.rb
```diff
@@ -34,7 +34,7 @@
 
   context 'when the step is `conviction_date` ' do
```
zheileman: This test is duplicated in lines 59:61

PR #69, opened 2019-06-12
Add Step: Enter length of conviction given in weeks or months?

tmckernan:
Separate conviction length and conviction length type into two separate questions.

### DIFF
--- a/app/controllers/steps/conviction/conviction_length_controller.rb
+++ b/app/controllers/steps/conviction/conviction_length_controller.rb
@@ -4,7 +4,6 @@ module Steps
       def edit
         @form_object = ConvictionLengthForm.new(
           disclosure_check: current_disclosure_check,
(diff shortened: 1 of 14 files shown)


app/controllers/steps/conviction/conviction_length_controller.rb
```diff
@@ -4,8 +4,7 @@ class ConvictionLengthController < Steps::ConvictionStepController
       def edit
         @form_object = ConvictionLengthForm.new(
           disclosure_check: current_disclosure_check,
-          conviction_length_type: current_disclosure_check.conviction_length_type,
-          conviction_length: current_disclosure_check.conviction_length
+          conviction_length_type: current_disclosure_check.conviction_length_type
```
zheileman: This should be `conviction_length`

PR #72, opened 2019-06-14
Add over 18 exit page

tmckernan:
Add exit page for over 18 and update the copy for caution condition exit.

Also remove general exit page.


zheileman:
If we are going to share the exit page between caution and conviction, then I think it should be in the `Check` namespace, so `/steps/check/exit_over18`

It feels weird entering the caution path and then exiting with a `/conviction/` in the URL.

Other than that 👍

app/controllers/steps/check/exit_over18_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Steps
+  module Check
+    class ExitOver18Controller < Steps::ConvictionStepController
```
zheileman:
```suggestion
    class ExitOver18Controller < Steps::CheckStepController
```

PR #78, opened 2019-06-21
Add conviction financial feature

tmckernan:
Add feature tests for conviction financial feature basis on [Conviction rules matrix](https://docs.google.com/spreadsheets/d/1ZSCk-wgMfIc22GQahKH_ys4u-E9GWIn6dQ7mm5t1RFI/edit#gid=0).

### DIFF
--- a/features/conviction_financial.feature
+++ b/features/conviction_financial.feature
@@ -7,6 +7,7 @@ Feature: Conviction
     When I choose "A fine"
     Then I should see "When did you get convicted?"
     When I enter a valid date
+    Then I should be on "/steps/check/results"
 
   @happy_path
   Scenario: Conviction Financial penalty - Compensation paid in full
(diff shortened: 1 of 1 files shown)


features/conviction_financial.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Conviction
+  Background:
+    When I am completing a basic under 18 "Financial penalty" conviction
+
+  @happy_path
+  Scenario: Conviction Financial penalty - Fine
+    When I choose "A fine"
+    Then I should see "When did you get convicted?"
+    When I enter a valid date
```
zheileman: Let's finish the scenarios with something that can "read" like a closure.
Otherwise it is like the scenario is missing something.

So in this case, after the `When I enter a valid date`
We can add another step like: `Then I should be on "/steps/check/results"`

So it is clear after entering the date the scenario is completed and no more questions are asked to the user.

Let's do this for all the scenarios please.

PR #79, opened 2019-06-21
Add conviction discharge feature

tmckernan:
Add feature tests for conviction discharge basis on [Conviction rules matrix](https://docs.google.com/spreadsheets/d/1ZSCk-wgMfIc22GQahKH_ys4u-E9GWIn6dQ7mm5t1RFI/edit#gid=0).

### DIFF
--- a/features/conviction_financial_discharge.feature
+++ b/features/conviction_financial_discharge.feature
@@ -8,17 +8,16 @@ Feature: Conviction
     And I choose "Absolute discharge"
     Then I should see "When did you get convicted?"
     When I enter a valid date
+    Then I should be on "/steps/check/results"
 
   @happy_path
   Scenario: Conviction Discharge - Conditional discharge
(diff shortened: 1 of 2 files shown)


features/conviction_financial_discharge.feature
```diff
@@ -0,0 +1,24 @@
...
+  Scenario: Conviction Discharge - Conditional discharge
+    And I choose "Conditional discharge"
+    Then I should see "When did you get convicted?"
+
+    When I enter a valid date
+
+    Then I should see "Was the length of conviction given in weeks, months or years?"
+    And I choose "Weeks"
+
+    Then I should see "What was the length of the order?"
```
zheileman: Here I would introduce an extra step to actually ensure we are picking the correct hint text based on the selection from before, if it was months, weeks or years.

It can be done as part of this same scenario or create a new scenario only for this.

PR #81, opened 2019-06-21
Add conviction youth rehabilitation order feature

tmckernan:
Add feature tests for conviction youth rehabilitation order feature basis on [Conviction rules matrix](https://docs.google.com/spreadsheets/d/1ZSCk-wgMfIc22GQahKH_ys4u-E9GWIn6dQ7mm5t1RFI/edit#gid=0).


features/conviction_youth_rehabilitation_order.feature
```diff
@@ -0,0 +1,63 @@
+Feature: Conviction
+  Background:
+    When I am completing a basic under 18 "Community or youth rehabilitation order (YRO)" conviction
+
+  @happy_path
+  Scenario: Conviction Community or youth rehabilitation order - Alcohol abstinence or treatment
+    When I am completing YRO with "Alcohol abstinence or treatment"
```
zheileman: I think we could refactor all these scenarios with an scenario outline (https://cucumber.io/docs/gherkin/reference/#scenario-outline) where the input is the conviction:

`I am completing YRO with <conviction>`

And the output or result would be:

`Then I should be on <end_page>` which, in all cases will be, at least for now, "/steps/check/results"

But it is up to you, maybe as an interesting exercise, if you are not too busy.

It just feel so weird these scenarios, they are like unfinished. We are now hiding too much in the step definitions and make it difficult to read and understand the scenarios so I think we need to strike a balance here.

tmckernan: I have moved the step definitions back into the  youth rehabilitation order feature tests and used scenario outline for each  sub type.

zheileman:
That is some work of art 😉 
Maybe now we can even get rid of the step definition `I am completing YRO with ...` if we don't intend to use in any other place.

PR #84, opened 2019-06-25
Add conviction end date calculator

tmckernan:
Small class to calculate the end of conviction by adding conviction start date (known_date) with conviction length.  i.e  2000-01-01 + 1.year. 

Please note:  conviction length type is saved as  plural (months, weeks, days) has to be singular for the calculation.

### DIFF
--- a/app/services/calculators/conviction_end_date_calculator.rb
+++ b/app/services/calculators/conviction_end_date_calculator.rb
@@ -1,0 +1,7 @@
+module Calculators
+  class ConvictionEndDateCalculator < BaseCalculator
+    def expiry_date
+      disclosure_check.known_date.advance(conviction_length)
+    end
+  end
+end
--- a/spec/services/conviction_expiry_calculator_spec.rb
+++ b/spec/services/conviction_expiry_calculator_spec.rb
@@ -1,14 +1,38 @@
 require 'rails_helper'
 
 RSpec.describe ConvictionExpiryCalculator do
-  subject { described_class.new(disclosure_check: disclosure_check) }
+  subject { described_class.new(disclosure_check: disclosure_check, ) }
 
   context '#expiry_date' do
-    let(:disclosure_check) { build(:disclosure_check, :conviction) }
-    let(:expiry_date) { subject.expiry_date }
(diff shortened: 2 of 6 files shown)


app/services/calculators/conviction_end_date_calculator.rb
```diff
@@ -0,0 +1,7 @@
+module Calculators
+  class ConvictionEndDateCalculator < BaseCalculator
+    def expiry_date
+      disclosure_check.known_date + disclosure_check.conviction_length.send(disclosure_check.conviction_length_type.singularize)
```
zheileman: We can do the following and that way avoid the `#send` method that always give me the creeps 😄 

```ruby
disclosure_check.known_date.advance(disclosure_check.conviction_length_type.to_sym => disclosure_check.conviction_length)
```

Ideally as we will be doing this operation many times, we can create in the superclass a method `#conviction_length` that returns the hash already: 

```ruby
def conviction_length
  { disclosure_check.conviction_length_type.to_sym => disclosure_check.conviction_length }
end
```

So we can just do this:

```ruby
disclosure_check.known_date.advance(conviction_length)
```

spec/services/conviction_expiry_calculator_spec.rb
```diff
@@ -1,14 +1,38 @@
...
   context '#expiry_date' do
-    let(:disclosure_check) { build(:disclosure_check, :conviction) }
+    let(:disclosure_check) { build(:disclosure_check, :conviction,
+                                   conviction_subtype: conviction_subtype,
+                                   known_date: known_date,
+                                   conviction_length: conviction_length,
+                                   conviction_length_type: conviction_length_type) }
+
     let(:expiry_date) { subject.expiry_date }
+    let(:known_date) { nil }
```
zheileman: Maybe we can try leaving the operations out of this spec, as otherwise we will end up having very similar or duplicated tests, the ones here, and the ones in each of the calculators.

The ideal would be to thoroughly test each of the individual 'calculators', and then ensure in this spec we call the proper calculator, but not really care about the results as the individual calculators should have test already that.

PR #90, opened 2019-06-26
Add immediately calculator

tmckernan:
Small class for when spent time is immediately.

Currently just return the start date of the conviction.

### DIFF
--- a/app/value_objects/conviction_type.rb
+++ b/app/value_objects/conviction_type.rb
@@ -38,7 +38,7 @@ class ConvictionType < ValueObject
     PROHIBITION                        = new(:prohibition,                      parent: COMMUNITY_ORDER, calculator_class: Calculators::YouthRehabilitationOrderCalculator),
     REFERRAL_ORDER                     = new(:referral_order,                   parent: COMMUNITY_ORDER, calculator_class: Calculators::YouthRehabilitationOrderCalculator),
     REHAB_ACTIVITY_REQUIREMENT         = new(:rehab_activity_requirement,       parent: COMMUNITY_ORDER, calculator_class: Calculators::YouthRehabilitationOrderCalculator),
-    REPARATION_ORDER                   = new(:reparation_order,                 parent: COMMUNITY_ORDER),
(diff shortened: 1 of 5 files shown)


spec/services/calculators/immediately_calculator_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+                                   conviction_length: conviction_length,
+                                   conviction_length_type: conviction_length_type) }
+
+    let(:known_date) { Date.new(2018, 10, 31) }
+    let(:conviction_length) { 10 }
+
+    context 'conviction length in months' do
+      let(:conviction_length_type) { 'months' }
+
+      it 'returns expiry date 10 months advance' do
```
zheileman: I think these tests doesn't correspond with this calculator. Probably copy paste.

app/value_objects/conviction_type.rb
```diff
@@ -38,7 +38,7 @@ def self.find_constant(raw_value)
     PROHIBITION                        = new(:prohibition,                      parent: COMMUNITY_ORDER, calculator_class: Calculators::YouthRehabilitationOrderCalculator),
     REFERRAL_ORDER                     = new(:referral_order,                   parent: COMMUNITY_ORDER, calculator_class: Calculators::YouthRehabilitationOrderCalculator),
     REHAB_ACTIVITY_REQUIREMENT         = new(:rehab_activity_requirement,       parent: COMMUNITY_ORDER, calculator_class: Calculators::YouthRehabilitationOrderCalculator),
-    REPARATION_ORDER                   = new(:reparation_order,                 parent: COMMUNITY_ORDER),
+    REPARATION_ORDER                   = new(:reparation_order,                 parent: COMMUNITY_ORDER, calculator_class: Calculators::ImmediatelyCalculator),
```
zheileman: This `reparation_order` conviction subtype should have the `skip_length: true` because is spent immediately (as the `absolute_discharge`). So we don't have to ask for the length. I think this was an oversight.

Then we can update the spec so the double doesn't need `conviction_length` or `conviction_length_type` methods. And maybe also cukes?

PR #91, opened 2019-06-27
Add fine calculator

tmckernan:
Small class to calculate the spent date for fine based on the following logic:
 - Conviction start date + 6 months

### DIFF
--- a/spec/services/calculators/fine_calculator_spec.rb
+++ b/spec/services/calculators/fine_calculator_spec.rb
@@ -2,31 +2,16 @@ require 'rails_helper'
 
 RSpec.describe Calculators::FineCalculator do
   subject { described_class.new(disclosure_check) }
-
-  context '#expiry_date' do
+   context '#expiry_date' do
     let(:disclosure_check) { build(:disclosure_check,
-                                   known_date: known_date,
-                                   conviction_length: conviction_length,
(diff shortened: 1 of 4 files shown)


spec/services/calculators/fine_calculator_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+RSpec.describe Calculators::FineCalculator do
+  subject { described_class.new(disclosure_check) }
+
+  context '#expiry_date' do
+    let(:disclosure_check) { build(:disclosure_check,
+                                   known_date: known_date,
+                                   conviction_length: conviction_length,
```
zheileman: We are skipping the length. These should not be in the double.

spec/services/calculators/fine_calculator_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+                                   conviction_length_type: conviction_length_type) }
+
+    let(:known_date) { Date.new(2016, 6, 20) }
+    let(:conviction_length) { 5 }
+
+    let(:result) { Date.new(2016, 12, 20) }
+
+    context 'conviction length in months' do
```
zheileman: These contexts doesn't correspond to this calculator...

PR #94, opened 2019-06-28
 Remove unused `condition_complied` step

zheileman:
Also remove the corresponding DB attribute, old locales, and the kickout page `condition_exit`.

This is done because the user flow has changed based on legal feedback and we no longer need to ask this question.

The decision tree and cukes have been updated accordingly.


spec/services/caution_decision_tree_spec.rb
```diff
@@ -4,68 +4,54 @@
   let(:disclosure_check) do
     instance_double(
       DisclosureCheck,
-      caution_type: caution_type,
-      condition_complied: condition_complied,
-      is_date_known: is_date_known,
```
tmckernan: if `is_date_known`  not used can we please remove it in another PR.

zheileman: Yes. I would do in a separate PR.
It has not been used for a while now.

PR #98, opened 2019-07-01
update conviction result page

tmckernan:
Update the conviction result page to return the correct title and statement in the right tense.

Copy can be [found here](https://docs.google.com/spreadsheets/d/1ZSCk-wgMfIc22GQahKH_ys4u-E9GWIn6dQ7mm5t1RFI/edit#gid=0)

### DIFF
--- a/app/helpers/conviction_helper.rb
+++ b/app/helpers/conviction_helper.rb
@@ -3,9 +3,7 @@ module ConvictionHelper
     return t('conviction_title.never', scope: 'results/conviction') unless object.instance_of?(Date)
 
     statement_type = object < Date.today ? 'past' : 'present'
-    statement = t("conviction_title.#{statement_type}", scope: 'results/conviction')
-
-    "#{statement} #{I18n.l(object)}"
+    t("conviction_title.#{statement_type}", scope: 'results/conviction', date: I18n.l(object))
   end
 
   def conviction_statement_string(object)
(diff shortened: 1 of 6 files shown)


app/helpers/conviction_helper.rb
```diff
@@ -0,0 +1,17 @@
+module ConvictionHelper
+  def conviction_title_string(object)
+    return t('conviction_title.never', scope: 'results/conviction') unless object.instance_of?(Date)
+
+    statement_type = object < Date.today ? 'past' : 'present'
+    statement = t("conviction_title.#{statement_type}", scope: 'results/conviction')
```
zheileman: We can pass the date here `date: I18n.l(object)` and then parametrise the locale (see comment below).

That way if we were to alter the order where the date appears in the sentence, (or even add a fullstop after the date) we don't need to touch the code here, just the locale.

config/locales/en/results.yml
```diff
@@ -68,7 +68,18 @@ en:
         youth_simple_caution: Youth Simple caution
         youth_conditional_caution: Youth Conditional caution
 
+
   results/conviction:
+    conviction_title:
+      past: Your conviction was spent on
```
zheileman:
```suggestion
      past: Your conviction was spent on %{date}
```

(same for `present` sentence)

PR #103, opened 2019-07-02
Update caution result to have the correct tense

tmckernan:
Update the caution result page to return the correct title and statement in the right tense.

Copy can be [found here](https://docs.google.com/spreadsheets/d/1ZSCk-wgMfIc22GQahKH_ys4u-E9GWIn6dQ7mm5t1RFI/edit#gid=0)

Allow conditional_end_date to accept future dates and update shared specs.


app/helpers/result_helper.rb
```diff
@@ -0,0 +1,17 @@
+module ResultHelper
+  def title_string(object, kind)
+    invalid_date = kind == 'conviction' ? 'never' : 'caution_result'
```
zheileman: I think if we get rid of this statement, the code will be simpler and have the same result. Naming the locale key to `no_date` will also be more explicit.

In the following line we can then just say:

`return t("title.no_date", scope: "results/#{kind}") unless object.instance_of?(Date)`

app/helpers/result_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+    return t("#{kind}_title.#{invalid_date}", scope: "results/#{kind}") unless object.instance_of?(Date)
+
+    title_type = object < Date.today ? 'past' : 'present'
+    t("#{kind}_title.#{title_type}", scope: "results/#{kind}", date: I18n.l(object))
```
zheileman: Following the above comment, here we don't really need to prefix the locale `"#{kind}_title` as we are already scoping them. So just `title.no_date` would be enough.

If we were to be super sure in case in the future we add an attribute to the DB named `title` or similar, we can extend the scope to read: `scope: "results/#{kind}"/confirmation_panel`

app/helpers/result_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def statement_string(object, kind)
```
zheileman: The same comments would apply to this other method.

spec/forms/steps/caution/conditional_end_date_form_spec.rb
```diff
@@ -1,5 +1,5 @@
 require 'spec_helper'
 
 RSpec.describe Steps::Caution::ConditionalEndDateForm do
-  it_behaves_like 'a date question form', attribute_name: :conditional_end_date
+  it_behaves_like 'a date question form', attribute_name: :conditional_end_date, future_date: true
```
zheileman: Can we name this to `allow_future`, so it is consistent with the intended validation.

PR #105, opened 2019-07-03
Add armed forces conviction type and subtypes

tmckernan:
Add Armed forces conviction type and subtype with the correct flow and result.  I am waiting on the correct copy.

### DIFF
--- a/config/locales/en/helpers.yml
+++ b/config/locales/en/helpers.yml
@@ -2,12 +2,18 @@
 en:
   dictionary:
     CONVICTION_TYPES: &CONVICTION_TYPES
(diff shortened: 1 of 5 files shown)


config/locales/en/helpers.yml
```diff
@@ -64,6 +70,8 @@ en:
         # default key
         known_date: When were you given the order?
         # alternative keys defined with `i18_attribute`
+        dismissal: When were you given the dismissal?
+        service_detention: When were you given the dismissal?
```
zheileman: This needs to be `When were you given the detention?` according to the spreadsheet.

PR #119, opened 2019-07-10
remove google tag manager and add google analytics

tmckernan:
Remove google tag manager code.

Add google analytics code.  

Now send, which radio button was selected when any form is submit i.e

Screen Shot 2019-07-10 at 15 05 38

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -10,13 +10,21 @@
 // Read Sprockets README (https://github.com/rails/sprockets#sprockets-directives) for details
 // about supported directives.
 //
-//= require rails-ujs
-//= require_tree .
-
 //= require jquery
+//= require rails-ujs
 // GOV.UK Frontend
 // https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md
 //
 //= require govuk-frontend/all
 
 //= require moj
+//= require_tree ./modules
+
+$(document).ready(function() {
+
+  // Initialize JS in /modules\
+  moj.init();
+});
+
+
+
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/module_loader.js
```diff
@@ -0,0 +1,36 @@
+(function() {
```
zheileman: We have already in `vendor/assets/javascripts` the file moj.js that serves this very same purpose. Why create a new one?

app/assets/javascripts/application.js
```diff
@@ -10,13 +10,20 @@
...
-//= require jquery
 // GOV.UK Frontend
 // https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md
 //
 //= require govuk-frontend/all
 
 //= require moj
+
+
+$(document).ready(function() {
```
zheileman: We are already requiring and initialising the modules in the above line `//= require moj`
It uses the already present `moj.js` file. Do we need again this?

app/views/layouts/_analytics.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+  ga('create', '<%= analytics_tracking_id %>', 'auto');
+
+  ga('set', 'anonymizeIp', true);
+  ga('send', 'pageview');
+
+  <% if content_for?(:transaction_data) %>
```
zheileman: Please remove this IF. It is not in the C100 anymore (I think you may have an old checked out copy locally).
We don't use this GA functionality so no need to have this code.

app/assets/javascripts/application.js
tmckernan: yes we need to initialise `moj`

PR #121, opened 2019-07-11
Add compensation_payment_date to conviction result table

tmckernan:
Add  compensation_payment_date to conviction result table

Screen Shot 2019-07-11 at 11 43 06


tmckernan:
will update the specs in another PR

PR #127, opened 2019-07-15
Add daily maintenance rake task

zheileman:
This will run daily through a cronjob pod, and will purge old checks so the database doesn't grow unnecessarily with stale data.

### DIFF
--- a/app/models/disclosure_check.rb
+++ b/app/models/disclosure_check.rb
@@ -3,4 +3,11 @@ class DisclosureCheck < ApplicationRecord
     in_progress: 0,
     completed: 10,
   }
+
+  # Using UUIDs as the record IDs. We can't trust sequential ordering by ID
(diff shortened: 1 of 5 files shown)


zheileman:
@tmckernan Precisely because of that, we would want to remove those checks early, as the data would be incomplete and not very useful for analytics, but still a few days is valuable to detect where users might be dropping out of the journey, or kickout (exit) pages. 7 days is a reasonable time period. Happy to go with any other (short) period.

Similarly for completed checks, and given there is no personal data in this service, we can go with a longer period, to be able to analyse the data and have enough checks to find patterns and/or iterate those checks that are most common, etc.
In the C100 we maintain these 30 days, but for this service we can go much longer.

tmckernan:
Fine lets go with these time frames for now

PR #132, opened 2019-07-16
Remove Hospital or guardianship order

tmckernan:
Remove Hospital or guardianship order conviction type

Merge custodial sentence conviction type with hospital order to create Custodial sentence or hospital order conviction type


app/value_objects/conviction_type.rb
```diff
@@ -19,11 +19,10 @@ def self.find_constant(raw_value)
 
   VALUES = [
     PARENT_TYPES = [
-      COMMUNITY_ORDER       = new(:community_order),
-      CUSTODIAL_SENTENCE    = new(:custodial_sentence),
-      DISCHARGE             = new(:discharge),
-      FINANCIAL             = new(:financial),
-      HOSPITAL_GUARD_ORDER  = new(:hospital_guard_order),
+      COMMUNITY_ORDER                    = new(:community_order),
+      CUSTODIAL_SENTENCE_HOSPITAL_ORDER  = new(:custodial_sentence_hospital_order),
```
zheileman: I wonder if we should still leave `CUSTODIAL_SENTENCE` as the main top category, after all, a quick search on google says: "A hospital order can also be used to confine a person to a psychiatric hospital after having been found not guilty of a crime by reason of insanity", so it is a custodial sentence after all, only not in prison.

That way we maintain the parent categories short and sweet 😄 
But this is just me! Feel free to leave it long if you like it.

PR #138, opened 2019-07-17
Bump to ruby 2.6.3 and rails 5.2.3

zheileman:
And also bumped many dependencies because of this.

No breaking changes. Mostly minor releases.

Also, forced by the bump in gems, I've replaced PhantomJS with headless Chrome. Individual commit with more context.


zheileman:
Opps, some issue with phantomJS on CircleCI.
Taking a look...

PR #142, opened 2019-07-23
Update content in conditional end date view

tmckernan:
Add expandable and collapsable box with content about when a condition has not ended.

Update hint text.

### DIFF
--- a/app/views/steps/caution/conditional_end_date/edit.html.erb
+++ b/app/views/steps/caution/conditional_end_date/edit.html.erb
@@ -11,6 +11,8 @@
 
         <%= step_form @form_object do |f| %>
           <%= f.gov_uk_date_field :conditional_end_date %>
+
+          <%= t ".conditions_not_ended_html" %>
           <%= f.continue_button %>
         <% end %>
       </div>
--- a/app/views/steps/caution/conditional_end_date/edit.html.erb
+++ b/app/views/steps/caution/conditional_end_date/edit.html.erb
@@ -12,18 +12,17 @@
         <%= step_form @form_object do |f| %>
           <%= f.gov_uk_date_field :conditional_end_date %>
 
+          <details class="govuk-details">
+            <summary class="govuk-details__summary">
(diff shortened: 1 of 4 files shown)


app/views/steps/caution/conditional_end_date/edit.html.erb
```diff
@@ -11,6 +11,8 @@
 
         <%= step_form @form_object do |f| %>
           <%= f.gov_uk_date_field :conditional_end_date %>
+
+          <%= t ".conditions_not_ended_html" %>
```
zheileman: Instead of dropping all the HTML in the locales, let's try to have as much markup as possible in the view itself, for example like this (but with the new design system markup and classes):

https://github.com/ministryofjustice/c100-application/blob/2ecc2de0da04c2219466c66011f61f628f1f4a8b/app/views/steps/application/urgent_hearing/edit.html.erb#L18

And then in the locales we only have the title and the content (this one, will contain some HTML, that's fine).

Also, let's add the `data-ga` attributes so we can track when users open/close the details. And change the class name in [ga-events.js](https://github.com/ministryofjustice/disclosure-checker/blob/master/app/assets/javascripts/modules/ga-events.js#L7) so it uses the new design system names.

config/locales/en/caution.yml
```diff
@@ -16,3 +16,14 @@ en:
       caution_type:
         edit:
           page_title: Caution types
+      conditional_end_date:
+        edit:
+          page_title: Caution conditional end date
+          conditions not ended_title: My conditions have not ended
```
zheileman: Key with spaces?

app/views/steps/caution/conditional_end_date/edit.html.erb
```diff
@@ -11,6 +11,19 @@
 
         <%= step_form @form_object do |f| %>
           <%= f.gov_uk_date_field :conditional_end_date %>
+
```
zheileman: A lot of extra space here.

PR #171, opened 2019-08-08
Parallelize linters and tests in CircleCI

zheileman:
Trying to speedup the time it takes to run the tests, specially now we have many cucumber features and these are slower than the specs.

Note: this is a bit of try and error, too many things could go wrong.


zheileman:
I'm spending more time on this than the hypothetical time we can save.
Closing PR and reverting to what we have.

PR #189, opened 2019-08-19
update start page and t&c copy

tmckernan:
Update start page and term and conditions copy 

The reason for doing the start page and t&c copy in one pull request because start page was linking to a new section in the t&c page.

Screen Shot 2019-08-19 at 12 04 12

### DIFF
--- a/app/views/about/terms_and_conditions.en.html.erb
+++ b/app/views/about/terms_and_conditions.en.html.erb
@@ -8,57 +8,48 @@
         <h1 class="govuk-heading-xl">Terms and conditions</h1>
         <p class="govuk-body">By using this service you agree to these terms and conditions.</p>
 
-
-        <section class="moj-Section about-section">
-          <h2 class="govuk-heading-m">About the service</h2>
-          <div class="Section__content govuk-govspeak gv-s-prose">
-            <p class="govuk-body">Check when a caution or conviction expires is a service provided by the Ministry of Justice (MoJ).</p>
-            <p class="govuk-body">Its purpose is to help users work out whether or not they need to tell organisations (such as employers, places of education or insurance companies) about a caution or conviction on their criminal record.</p>
-            <p class="govuk-body">The service gives users guidance on when a single caution or conviction expires (becomes ‘spent’) according to the current law in England and Wales.</p>
-            <div class="govuk-inset-text">
-              If you have more than one conviction, the guidance the service gives may not apply to you.
-            </div>
-            <p class="govuk-body">If you have been given another conviction during your rehabilitation period, or if you are applying for certain sensitive jobs, the guidance will not apply.</p>
-          </div
-        </section>
-
-        <section class="moj-Section about-section">
-          <h2 class="govuk-heading-m">How the service works</h2>
-          <div class="Section__content govuk-govspeak gv-s-prose">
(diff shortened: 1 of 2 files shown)


app/views/about/terms_and_conditions.en.html.erb
```diff
@@ -4,52 +4,62 @@
...
-            <p class="govuk-body">The Ministry of Justice (MOJ) runs this service.</p>
-            <p class="govuk-body">We try to make sure the information we provide is accurate and complete and the service is always available and virus free.</p>
-            <p class="govuk-body">The MoJ accepts no liability for:</p>
-            <ul class="govuk-list govuk-list--bullet">
-              <li>you being unable to access the service (for example due to high demand or because we've taken it down for maintenance)</li>
-              <li>indirect damages arising from you using the service or relying on its content</li>
-              <li>the content and availability of external websites that link to the service or which link from it</li>
-            </ul>
-          </div>
+            <p class="govuk-body">Check when a caution or conviction expires is a service provided by the Ministry of Justice (MoJ).</p>
```
zheileman: Let's use instead of the hardcoded service name, the helper (as we did in the cookies page for example) so if we ever change the name there is only one place that needs to be changed, `<%= service_name -%>`

app/views/about/terms_and_conditions.en.html.erb
```diff
@@ -4,52 +4,62 @@
...
           <div class="Section__content govuk-govspeak gv-s-prose">
-            <p class="govuk-body">We collect information about you in accordance with our <%= link_to 'privacy policy', about_privacy_path, class: 'govuk-link ga-pageLink', data: {ga_category: 'footer', ga_label: 'privacy'} %> and our <%= link_to 'cookie policy', about_cookies_path, class: 'govuk-link ga-pageLink', data: {ga_category: 'footer', ga_label: 'cookies'} %>. By using this service you agree to us collecting this information and confirm that any data you provide is accurate.</p>
-          </div>
+            <p class="govuk-body">As the service does not ask for specific details about users’ criminal records, it can only give general guidance on when cautions or convictions become spent in England and Wales.</p>
+
+            <div class="govuk-inset-text">
+               <p class="govuk-body">If you will be using the guidance given by this service to decide whether or not to disclose a caution or conviction on your criminal record, you are strongly advised to consider getting specific professional or legal advice before doing so.</p>
+            </div>
+          </div
```
zheileman: This `div` is not closed.

app/views/about/terms_and_conditions.en.html.erb
```diff
@@ -4,52 +4,62 @@
...
-            </p>
-          </div>
+            <p class="govuk-body">The expiry date given at the end of the service is based solely on the information provided by the service user.</p>
+            <p class="govuk-body">If any of the information a user inputs is incorrect, the date given at the end might also be incorrect.</p>
+            <p class="govuk-body">MoJ is not responsible for:</p>
+             <ul class="govuk-list govuk-list--bullet">
+              <li>people disclosing their expired cautions or convictions when they weren’t required to</li>
+              <li>how and when employers ask for information about people’s criminal records</li>
+            </ul>
+          </div
```
zheileman: This `div` is not closed.

PR #193, opened 2019-08-19
update start page copy

tmckernan:
Need to change "how to use the guidance this service gives you" bullet to:

"how to use the information this service gives you"

(and same change to header in T&Cs).

Start page change
Screen Shot 2019-08-19 at 14 33 00

T&C change
Screen Shot 2019-08-19 at 14 32 47

### DIFF
--- a/app/views/home/index.en.html.erb
+++ b/app/views/home/index.en.html.erb
@@ -23,7 +23,7 @@
         <ul class="govuk-list govuk-list--bullet">
           <li>which type of caution or conviction you were given</li>
           <li>the date your caution or conviction began</li>
-          <li><%= link_to 'how to use the information this service gives you', about_terms_and_conditions_path(anchor: "using-guidance"), class: 'govuk-link ga-pageLink', data: {ga_category: 'home', ga_label: 'how to use the information this service gives you'} %></li>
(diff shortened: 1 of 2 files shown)


app/views/home/index.en.html.erb
```diff
@@ -23,7 +23,7 @@
         <ul class="govuk-list govuk-list--bullet">
           <li>which type of caution or conviction you were given</li>
           <li>the date your caution or conviction began</li>
-          <li><%= link_to 'how to use the guidance this service gives you', about_terms_and_conditions_path(anchor: "using-guidance"), class: 'govuk-link ga-pageLink', data: {ga_category: 'home', ga_label: 'how to use the guidance this service gives you'} %></li>
+          <li><%= link_to 'how to use the information this service gives you', about_terms_and_conditions_path(anchor: "using-guidance"), class: 'govuk-link ga-pageLink', data: {ga_category: 'home', ga_label: 'how to use the information this service gives you'} %></li>
```
zheileman: Let's not attach the GA label to whatever is the copy here as if it starts changing (like now), our GA will be a mess. The label is kind of a unique ID that should not change (or rarely change).

Let's pick something abstract enough, like for example:

`ga_label: 'terms and conditions guidance'`

We could also go just with 'terms and conditions' to be fair, as there is only one link to terms and conditions in the Home (and another in the Footer) and we are already setting the ga_category to `home`.

PR #194, opened 2019-08-19
update question copy for DTO length question

tmckernan:
Screen Shot 2019-08-19 at 15 02 44


tmckernan:
Screen Shot 2019-08-19 at 15 38 42

PR #229, opened 2019-09-16
Remove/cleanup the opt-in/opt-out mechanism

tmckernan:
Remove the following:

- info form object
- mvp decision tree
- info confirmation page
- all related specs

Rename database column opt-in to opt-out

Update landing page copy


app/views/steps/mvp/info/edit.en.html.erb
```diff
@@ -12,8 +12,7 @@
         </h1>
 
         <p class="govuk-body">
-          Over the past few months, the User-Centred Policy Design (UCPD) team at the Ministry of Justice has been
-          looking at ways to improve the information and guidance available to young offenders.
+          Over the past few months, the User-Centred Policy Design (UCPD) team at the Ministry of Justice has been looking at ways to improve the information and guidance available to young offenders.
```
zheileman: There are many changes in this file that are not really changes just break lines.

This makes for a very difficult PR review because I have to read every single line trying to look for the change. Also mess completely the history.

Can you please only modify the line or lines that genuinely contains a real copy change?

I take you copy&paste'd whole lines, but as you can see, that produces fake changes.
We should be reading what we copy&paste not only to not change things that are not really changes, but also more importantly to not introduce typos or errors inadvertently.

app/views/steps/mvp/info/edit.en.html.erb
```diff
@@ -118,14 +116,23 @@
...
+
+         <p class="govuk-body">
+          If you choose to take part in our pilot, please let any young people know if the spent date that’s given at the end of the service is different to the date that you would have given them, and send us feedback that the result is incorrect
         </p>
 
-        <%= step_form @form_object, html: {class: 'govuk-!-margin-top-8'} do |f| %>
-          <%= f.radio_button_fieldset :opted_in, choices: GenericYesNo.values,
-                                      legend_options: {page_heading: false, visually_hidden: false, size: 'm'} %>
-          <%= f.continue_button %>
+        <%= link_to service_name, class: 'govuk-button govuk-button--start govuk-!-margin-top-2', data: { module: 'govuk-button' }, role: 'button', draggable: false do %>
```
zheileman: Have you tried this button?

PR #230, opened 2019-09-16
Move the landing page outside the steps framework

tmckernan:
Move landing page outside the step framework as there isn't any reference to step view/controller etc.

Move the URL from `https://{host}/steps/mvp/info/{reference`  to `https://{host}/pilot/{reference}`


app/controllers/pilot_controller.rb
```diff
@@ -0,0 +1,16 @@
+class PilotController < ApplicationController
+  before_action :validate_reference
```
zheileman: I think we will still want this page to be accessible without credentials:

```ruby
skip_before_action :check_http_credentials
```

PR #246, opened 2019-09-26
Update adult motoring journey to include conviction types

tmckernan:
Add motoring endorsement step after all the adult motoring sub conviction types

### DIFF
--- a/app/decorators/conviction_decorator.rb
+++ b/app/decorators/conviction_decorator.rb
@@ -12,4 +12,13 @@ module ConvictionDecorator
       ConvictionType::ADULT_CUSTODIAL_SENTENCE,
     ].include?(self)
   end
+
+  def motoring?
+    [
+      ConvictionType::ADULT_DISQUALIFICATION,
+      ConvictionType::ADULT_MOTORING_FINE,
+      ConvictionType::ADULT_PENALTY_NOTICE,
+      ConvictionType::ADULT_PENALTY_POINTS
+    ].include?(self)
+  end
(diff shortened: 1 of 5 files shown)


app/decorators/conviction_decorator.rb
```diff
@@ -12,4 +12,13 @@ def custodial_sentence?
       ConvictionType::ADULT_CUSTODIAL_SENTENCE,
     ].include?(self)
   end
+
+  def motoring?
```
zheileman: Please use the inquirer as per example in my last PR. I specifically use motoring as the example.

No need to create any new method.
All motoring convictions are part of parent group, so inquiring the parent will give you what you need.

app/services/conviction_decision_tree.rb
```diff
@@ -28,6 +30,7 @@ def destination
 
   def after_conviction_subtype
     return edit(:compensation_paid) if conviction_subtype.compensation?
+    return edit(:motoring_endorsement) if conviction_subtype.motoring?
```
zheileman: Use the enquirer.

spec/decorators/conviction_decorator_spec.rb
```diff
@@ -29,4 +29,31 @@
       it { expect(subject.custodial_sentence?).to eq(false) }
     end
   end
+
+  describe '#motoring?' do
```
zheileman: Then we don't need these tests (as these will be tested already via proxy the decision tree).

PR #256, opened 2019-10-02
Update adult penalty notice journey

tmckernan:
- Add new non conviction results page
- Exit adult penalty notice journey if the conviction isn't endorsed

### DIFF
--- a/spec/services/conviction_decision_tree_spec.rb
+++ b/spec/services/conviction_decision_tree_spec.rb
@@ -131,24 +131,32 @@ RSpec.describe ConvictionDecisionTree do
   end
 
   context 'when the step is `motoring_endorsement`' do
-    let(:conviction_subtype) { :adult_penalty_notice }
     let(:motoring_endorsement) {GenericYesNo::YES }
     let(:step_params) { { motoring_endorsement:  motoring_endorsement } }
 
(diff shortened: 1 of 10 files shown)


spec/services/conviction_decision_tree_spec.rb
```diff
@@ -129,8 +131,25 @@
...
 
   context 'when the step is `motoring_endorsement`' do
-    let(:step_params) { { motoring_endorsement: 'anything' } }
-    it { is_expected.to have_destination(:known_date, :edit) }
+    let(:conviction_subtype) { :adult_penalty_notice }
+    let(:motoring_endorsement) {GenericYesNo::YES }
+    let(:step_params) { { motoring_endorsement:  motoring_endorsement } }
+
+    context 'when subtype equal adult_penalty_notice with a endorsement' do
+      let(:conviction_subtype) { :adult_penalty_notice }
```
digitalronin: You don't need line 139, because you've got line 134 in this context.

But, for clarity, it might be worth using sub-contexts "when subtype equal adult_penalty_notice" and "when subtype equal adult_disqualification".

Another option is to move line 134 here, and copy it into the other context, too. 

Either way, specifying the conviction_subtype would be consistent for each context.

PR #257, opened 2019-10-04
Allow no end date for motoring disqualification

tmckernan:
Allow a user not to enter a date in the motoring disqualification end date step if they don't have an end date.

### DIFF
--- a/app/services/calculators/motoring_calculator.rb
+++ b/app/services/calculators/motoring_calculator.rb
@@ -4,30 +4,40 @@ module Calculators
 
     # If a lifetime ban was given:
     #  - never spent
+    # If no end_date was given: Start date + 5 years
+    #  - Start date + 5 years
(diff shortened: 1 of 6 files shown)


app/services/calculators/motoring_calculator.rb
```diff
@@ -4,30 +4,40 @@ class MotoringCalculator < BaseCalculator
...
+    #  - Start date + 5 years
     # If an endorsement was received
     #  - If (end_date - start_date) is less than or equal to 5 years: start_date + 5 years
     #  - If (end_date - start_date) is greater than 5 years: end_date
 
     # If an endorsement was not received
-    #  - end date
+    # If no end_date was given:
+    #  - Start date + 2 years
+    # with and End date
```
digitalronin: typo: with and -> with an

PR #258, opened 2019-10-07
Update result page content

tmckernan:
Several copy changes to the final result page.


zheileman:
@tmckernan I've noticed in this PR there are copy changes in the templates:

* show.en.html+conviction_spent.erb
* show.en.html+conviction_not_spent.erb

Should this also be applied to the following templates?

* show.en.html+conviction_never_spent.erb
* show.en.html+caution_spent.erb
* show.en.html+caution_not_spent.erb

Because the disclaimer also applies to them.
Maybe we can extract it to a partial to be shared in all the templates as we do with other blocks.

PR #271, opened 2019-10-18
Display multiple check on check your answer page

tmckernan:
Basic display multiple disclosures question and answers.  At the moment it display all checks as a separate entity.  

Screen Shot 2019-10-18 at 12 40 56

### DIFF
--- a/app/controllers/steps/check/check_your_answers_controller.rb
+++ b/app/controllers/steps/check/check_your_answers_controller.rb
@@ -2,7 +2,7 @@ module Steps
   module Check
     class CheckYourAnswersController < Steps::CheckStepController
       def show
-        @presenters = CheckAnswersPresenter.new(current_disclosure_report)
+        @presenter = CheckAnswersPresenter.new(current_disclosure_report)
       end
     end
   end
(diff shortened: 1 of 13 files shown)


app/controllers/steps/check/check_your_answers_controller.rb
```diff
@@ -2,7 +2,11 @@ module Steps
   module Check
     class CheckYourAnswersController < Steps::CheckStepController
       def show
-        @presenter = ResultsPresenter.build(current_disclosure_check)
+        @presenters = current_disclosure_report.check_groups.map do |check_group|
```
zheileman: I suggest creating a separate presenter, named something like CheckAnswersPresenter (or similar) that receives the `current_disclosure_report` and encapsulates all this logic.

That way you still return a single presenter (not a collection of presenters) and is this presenter that knows how to present itself and do the loops etc.

Ideally we would end up with a very similar scenario to what we have now, where:

```html
<dl class="govuk-summary-list">
   <%= render @presenter.summary %>
</dl>
```

Where `@presenter` is the above mentioned `CheckAnswersPresenter`.

Otherwise we risk ending up with too much complexity in the views, and most importantly, not a single spec. The presenter must be tested separately to see if it is doing what is supposed to do.

config/locales/en/results.yml
```diff
@@ -79,6 +79,9 @@ en:
       adult_suspended_prison_sentence: Suspended prison sentence
       adult_prison_sentence: Prison sentence
 
+  caution: Caution
```
zheileman: Let's namespace this, so:

```yaml
kind:
  caution: Caution
  conviction: Conviction
```

PR #284, opened 2019-10-25
Add Accessibility statement

tmckernan:
- Add link in the footer to accessibility statement
- Add  accessibility statement copy

You can find the copy document is [here](https://docs.google.com/document/d/1x1fsCFG2ewlixVFlZhKQ-HxSR2jYoIO1V7Ryj96fJIQ/edit#heading=h.tbi70ua5oje3).

### DIFF
--- a/app/views/about/accessibility.en.html.erb
+++ b/app/views/about/accessibility.en.html.erb
@@ -79,7 +79,7 @@
 
         <p class="govuk-body">This statement was prepared on 24 October 2019.</p>
 
-        <p class="govuk-body">It was last updated on 24 October 2019.</p>
+        <p class="govuk-body">It was last updated on 28 October 2019.</p>
       </div>
     </div>
   </main>
--- a/app/views/about/accessibility.en.html.erb
+++ b/app/views/about/accessibility.en.html.erb
@@ -1,0 +1,84 @@
+<% title 'Accessibility statement for Disclosure Checker' %>
+<div class="govuk-width-container">
+  <main class="govuk-main-wrapper"  id="main-content">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-xl">Accessibility statement for Disclosure Checker</h1>
+
+        <h2 class="govuk-heading-m">Using this online tool</h2>
+
+        <p class="govuk-body">This online tool is run by the Ministry of Justice. We want as many people as possible to be able to use it.</p>
+
+        <p class="govuk-body">For example, that means you should be able to:</p>
+
+        <ul class="govuk-list govuk-list--bullet">
+          <li>change colours, contrast levels and fonts</li>
+          <li>zoom in up to 300% without the text spilling off the screen</li>
+          <li>navigate most of the online tool using just a keyboard</li>
+          <li>navigate most of the online tool using speech recognition software</li>
+        </ul>
+
+        <p class="govuk-body">We've also made the text as simple as possible to understand.</p>
+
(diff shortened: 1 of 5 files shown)


config/locales/en/layouts.yml
```diff
@@ -7,6 +7,7 @@ en:
       cookies: Cookies
       privacy: Privacy
       terms_and_conditions: Terms and conditions
+      accessibility: Accessibility statement
```
zheileman: I would name this just 'Accessibility' for short, in line with other services I've seen, like the Claim for costs of a child’s funeral.

app/views/about/accessibility.en.html.erb
```diff
@@ -0,0 +1,84 @@
...
+
+        <h2 class="govuk-heading-m">Reporting accessibility problems</h2>
+
+        <p class="govuk-body">We're always looking to improve the accessibility of this online tool. If you find any problems not listed on this page or you think we're not meeting accessibility requirements, contact <a href="mailto:disclosure.checker@digital.justice.gov.uk">disclosure.checker@digital.justice.gov.uk</a>.</p>
+
+        <h2 class="govuk-heading-m">Enforcement procedure</h2>
+
+        <p class="govuk-body">The Equality and Human Rights Commission (EHRC) is responsible for enforcing the Public Sector Bodies (Websites and Mobile Applications) (No. 2) Accessibility Regulations 2018 (the ‘accessibility regulations'). If you're not happy with how we respond to your complaint, <a class="govuk-link" href="https://www.equalityadvisoryservice.com/" rel="external" target="_blank">contact the Equality Advisory and Support Service (EASS)</a>.</p>
+
+        <p class="govuk-body">This statement was prepared on 24 October 2019 It was last updated on 24 October 2019.</p>
```
zheileman: I think a full stop or a new line will be nice here, I've added a comment in Wes' document although is a minor thing.

app/views/about/accessibility.en.html.erb
```diff
@@ -0,0 +1,84 @@
+<% title 'Accessibility statement for Disclosure Checker' %>
+<div class="govuk-width-container">
+  <main class="govuk-main-wrapper"  id="main-content">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-xl">Accessibility statement for Disclosure Checker</h1>
```
zheileman: I don't think we should publicly name this `Disclosure Checker`, that is more like an internal code-name, and instead use the name we have so far for the service `Check when a caution or conviction is spent` (in fact, the helper) as we do in the cookies or terms and conditions, `<%= service_name -%>`.
There are a few occurrences of this we can replace, including the title.

I've added a comment in Wes' document for clarification.

app/views/about/accessibility.en.html.erb
```diff
@@ -0,0 +1,84 @@
...
+        <ul class="govuk-list govuk-list--bullet">
+          <li>What type of conviction did you get?</li>
+          <li>When did the detention start?</li>
+        </ul>
+
+        <p class="govuk-body">Finally, we tested cookies and the third-party supplier feedback survey page.</p>
+
+        <h2 class="govuk-heading-m">What to do if you cannot access parts of this online tool</h2>
+
+        <p class="govuk-body">If you need information on this online tool in a different format like an accessible PDF, large print, easy read, audio recording or braille, email <a href="mailto:disclosure.checker@digital.justice.gov.uk">disclosure.checker@digital.justice.gov.uk</a>.</p>
```
zheileman: We can also add the `class="govuk-link"` to the `mailto` links (there are two).

PR #286, opened 2019-10-28
Only display motoring result copy for motoring convictions

tmckernan:
Display motoring copy for motoring convictions which have been  spent  or not spent.

Don't display motoring copy for life time bans i.e never spent motoring convictions.

### DIFF
--- a/app/presenters/conviction_result_presenter.rb
+++ b/app/presenters/conviction_result_presenter.rb
@@ -3,10 +3,6 @@ class ConvictionResultPresenter < ResultsPresenter
     'results/conviction'
   end
 
-  def motoring?
-    disclosure_check.conviction_type.inquiry.adult_motoring?
-  end
-
   private
 
   def question_attributes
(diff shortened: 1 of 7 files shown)


app/presenters/conviction_result_presenter.rb
```diff
@@ -3,6 +3,10 @@ def to_partial_path
     'results/conviction'
   end
 
+  def motoring?
```
zheileman: Let's move this to the decorator that was created for these situations, `ConvictionDecorator`, with a method along the lines:

```ruby
  def motoring?
    ConvictionType::ADULT_MOTORING.eql?(self)
  end
```

Then in the view, we use: `@presenter.conviction_type.motoring?`

app/presenters/conviction_result_presenter.rb
```diff
@@ -1,4 +1,6 @@
 class ConvictionResultPresenter < ResultsPresenter
+  include ConvictionDecorator
```
zheileman: This is not needed Tony as you don't access through the presenter, but instead through the `conviction_type`, which is the one already including the decorator.

PR #339, opened 2020-03-30
Add questions about compensation orders

Davidslv:
We are introducing two new questions around payments

### Diagram

Screenshot 2020-03-30 at 11 20 22

### Pages

Screenshot 2020-03-30 at 17 54 42

image

Screenshot 2020-03-30 at 17 54 28

Story: https://mojdigital.teamwork.com/#/tasks/19826007

### DIFF
--- a/app/controllers/steps/conviction/compensation_paid_amount_controller.rb
+++ b/app/controllers/steps/conviction/compensation_paid_amount_controller.rb
@@ -1,0 +1,13 @@
+module Steps
+  module Conviction
+    class CompensationPaidAmountController < Steps::ConvictionStepController
+      def edit
+        @form_object = CompensationPaidAmountForm.build(current_disclosure_check)
+      end
+
+      def update
+        update_and_advance(CompensationPaidAmountForm)
+      end
+    end
+  end
+end
--- a/app/services/conviction_decision_tree.rb
+++ b/app/services/conviction_decision_tree.rb
@@ -3,6 +3,7 @@ class ConvictionDecisionTree < BaseDecisionTree
 
   # rubocop:disable Metrics/CyclomaticComplexity
   # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize
   def destination
     return next_step if next_step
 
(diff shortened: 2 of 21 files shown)


app/controllers/steps/conviction/compensation_paid_amount_controller.rb
```diff
@@ -0,0 +1,16 @@
+module Steps
+  module Conviction
+    class CompensationPaidAmountController < Steps::ConvictionStepController
+      def edit
+        @form_object = CompensationPaidAmountForm.new(
```
zheileman: If you want you can probably replace these lines with a call to `build(current_disclosure_check)` like in the other controller (CompensationPaymentReceiptController).
But they behave the same, they are equivalent in this case.

app/views/steps/conviction/compensation_payment_receipt/edit.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <%= error_summary %>
+        <%= step_subsection %>
+
+        <!-- The header is part of the radios legend by default, but can be configured -->
+
+        <h1 class="govuk-heading-xl"><%=t '.heading' %></h1>
+
+        <div class="govuk-hint">
```
zheileman: You can probably set the hint through the locales and, like in the other view (compensation_paid_amount), set the legend to be also the heading, removing the `<h1>` too.
Take a look at the `hint` section in the helpers.yml file.

db/schema.rb
```diff
@@ -49,6 +49,8 @@
     t.boolean "approximate_conditional_end_date", default: false
     t.boolean "approximate_compensation_payment_date", default: false
     t.boolean "approximate_motoring_disqualification_end_date", default: false
+    t.string "compensation_payment_over_100"
+    t.string "sent_payment_receipt"
```
zheileman: Maybe we can rename this attribute to start also with compensation, as we try for all attributes related to some specific category to share part of the name so it's easy to refer and identify them.
Perhaps something like `compensation_receipt_sent` or similar. It's up to you!

app/services/conviction_decision_tree.rb
```diff
@@ -76,11 +84,27 @@ def after_conviction_length_type
...
 
   def after_compensation_paid
-    return edit(:compensation_payment_date) if GenericYesNo.new(disclosure_check.compensation_paid).yes?
+    return edit(:compensation_paid_amount) if GenericYesNo.new(disclosure_check.compensation_paid).yes?
 
     show(:compensation_not_paid)
   end
 
+  def after_compensation_paid_amount
+    edit(:compensation_payment_date)
```
zheileman: As mentioned in slack, in this case, you can just get rid of this method and add the `edit(:compensation_payment_date)` directly inside the case branch. We do that usually when the steps are lineal, with no conditions when moving from one to another.

PR #346, opened 2020-04-07
Update content on conviction type and community

Davidslv:
Story: https://mojdigital.teamwork.com/#/files/4650265

image

image

### DIFF
--- a/config/locales/en/helpers.yml
+++ b/config/locales/en/helpers.yml
@@ -246,7 +246,7 @@ en:
           military: For example, a dismissal or service detention
           prevention_reparation: For example, a restraining order or sexual harm prevention order
           # adults
-          adult_community_reparation: For example, a curfew, unpaid work or a restraining order. You might have been asked to wear a tag as part of your order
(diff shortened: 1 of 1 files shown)


config/locales/en/helpers.yml
```diff
@@ -246,7 +246,7 @@ en:
           military: For example, a dismissal or service detention
           prevention_reparation: For example, a restraining order or sexual harm prevention order
           # adults
-          adult_community_reparation: For example, a curfew, unpaid work or a restraining order. You might have been asked to wear a tag as part of your order
+          adult_community_reparation: For example, a sexual prevention order or a restraining order. You might have been asked to wear a tag as part of your order
```
zheileman, line 249: Probably better to use the complete "legal" term here, as we do already for under 18s: "sexual harm prevention order", just to be consistent.
https://hub.unlock.org.uk/knowledgebase/sexual-harm-prevention-order-shpo/

zheileman: Hmm, I've seen we use in the radios just Sexual order. So maybe it is ok to leave it just like that in the hint text too. We seem to have a mixture of terms though, would be good to standardised them.

Screen Shot 2020-04-07 at 09 06 39

PR #350, opened 2020-05-21
Introduce under age motoring convictions

Davidslv:
This commit introduces motoring convictions for under age, it slightly differs from adult motoring convictions,
for under age, the convictions are spent sooner.

The differences are:
- [Endorsement is 2.5 years (30 months)](https://www.gov.uk/guidance/telling-people-about-your-criminal-record#motoring) instead of 5 years (60 months)
- [Fines are 6 months](https://www.gov.uk/guidance/telling-people-about-your-criminal-record#fine) instead of 1 year (12 months)

Story: https://mojdigital.teamwork.com/#/tasks/20007708

### Example

image

### DIFF
--- a/app/services/conviction_decision_tree.rb
+++ b/app/services/conviction_decision_tree.rb
@@ -47,13 +47,13 @@ class ConvictionDecisionTree < BaseDecisionTree
   def after_conviction_subtype
     return edit(:conviction_bail)   if conviction_subtype.bailable_offense? && !step_name.eql?(:bypass_bail_conviction_subtype)
     return edit(:compensation_paid) if conviction_subtype.compensation?
-    return after_adult_motoring     if conviction_subtype.parent.inquiry.adult_motoring?
+    return after_adult_motoring     if conviction_subtype.parent.inquiry.adult_motoring? || conviction_subtype.parent.inquiry.youth_motoring?
 
     known_date_question
   end
 
   def after_adult_motoring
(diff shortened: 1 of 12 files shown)


app/services/conviction_decision_tree.rb
```diff
@@ -47,7 +47,7 @@ def destination
   def after_conviction_subtype
     return edit(:conviction_bail)   if conviction_subtype.bailable_offense? && !step_name.eql?(:bypass_bail_conviction_subtype)
     return edit(:compensation_paid) if conviction_subtype.compensation?
-    return after_adult_motoring     if conviction_subtype.parent.inquiry.adult_motoring?
+    return after_adult_motoring     if conviction_subtype.parent.inquiry.adult_motoring? || conviction_subtype.parent.inquiry.youth_motoring?
```
zheileman, line 50: Here you can replace the two conditions with the method from the decorator `#conviction_subtype.motoring?`

Apart from that, this decision tree is coupled in a few places to adult motoring, as you can see by the naming `return after_adult_motoring` so it needs a bit of refactoring to make either generic so it works for both kind of motoring, or make sure where there are differences between youth and adult, these are correct.

Cukes are failing because of this.

zheileman:
Cool, let's merge this for now as discussed, in order to start early testing on staging, with a view to improve it with some refactoring of the decision tree and the calculator in a coming PR, before releasing this to production in the future 👍

PR #355, opened 2020-05-28
Refactor conviction decision tree

Davidslv:
This is just a refactor of the latest changes done around motoring (adult and youth) as this was clogging the ConvictionDecisionTree.

The ConvictionDecorator is now the place where logic decisions are made between adult/youth.

### DIFF
--- a/app/services/conviction_decision_tree.rb
+++ b/app/services/conviction_decision_tree.rb
@@ -53,7 +53,7 @@ class ConvictionDecisionTree < BaseDecisionTree
   end
 
   def after_adult_motoring
-    return edit(:motoring_lifetime_ban) if conviction_subtype.inquiry.adult_disqualification? || conviction_subtype.inquiry.youth_disqualification?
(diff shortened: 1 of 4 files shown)


app/services/conviction_decision_tree.rb
```diff
@@ -47,13 +47,13 @@ def destination
   def after_conviction_subtype
     return edit(:conviction_bail)   if conviction_subtype.bailable_offense? && !step_name.eql?(:bypass_bail_conviction_subtype)
     return edit(:compensation_paid) if conviction_subtype.compensation?
-    return after_adult_motoring     if conviction_subtype.parent.inquiry.adult_motoring? || conviction_subtype.parent.inquiry.youth_motoring?
+    return after_adult_motoring     if conviction_subtype.parent.motoring?
```
zheileman, line 50: This can be just `conviction_type.motoring?`

Also, just to be coherent, what if we change the method name from `#after_adult_motoring` to something more generic, kind of `#after_motoring_conviction`?

Davidslv:
@zheileman I've addressed your comments, thank you!

PR #361, opened 2020-07-06
Remove life time ban from motoring disqualification

Davidslv:
We interpreted this as there being a difference between a disqualification period with no end date (where we applied the 2 year calculation) and a lifetime driving ban (if users selected this, we provided them with the information that their offence would never be spent).

We have now checked this with policy and legal, who have confirmed disqualification period with no end date and lifetime driving ban are the same thing, and a 2 year rehabilitation period applies.

Story: https://mojdigital.teamwork.com/#/tasks/20951241

@zheileman I've left the migration to remove the column to be done in a separate PR, let me know if you want it all together in this one.


features/youth/conviction_motoring.feature
```diff
@@ -52,7 +40,7 @@ Feature: Youth Conviction
     Examples:
       | subtype                    | endorsement | known_date_header                        | result               | spent_date                                      |
       | Fine                       | Yes         | When were you given the fine?            | /steps/check/results | This conviction will be spent on 1 July 2022    |
-      | Fine                       | No          | When were you given the fine?            | /steps/check/results | This conviction will be spent on 1 July 2020    |
+      | Fine                       | No          | When were you given the fine?            | /steps/check/results | This conviction was spent on 1 July 2020        |
```
zheileman, line 43: We need to freeze the time when running the cukes so the calculations are valid no matter when we run the tests.
But that can be done as a separate PR/ticket.

zheileman: https://github.com/ministryofjustice/disclosure-checker/pull/361/commits/184e8bb3c5f8b039a43b5e22744bdcd03d22d0f4

PR #362, opened 2020-07-15
Simplify journey for simple youth/adult cautions

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/21186055

- Update caution decision tree
- Update caution result presenter

image

Co-authored-by: @njseeto

### DIFF
--- a/app/presenters/caution_result_presenter.rb
+++ b/app/presenters/caution_result_presenter.rb
@@ -3,6 +3,19 @@ class CautionResultPresenter < ResultsPresenter
     'results/caution'
   end
 
+  def variant
+    return 'caution_simple' if caution_type.simple?
+
+    tense = if expiry_date.instance_of?(Date)
+              expiry_date.past? ? :spent : :not_spent
+            else
+              expiry_date
+            end
+
+    # The tense can be one of these values: spent, not_spent, never_spent or no_record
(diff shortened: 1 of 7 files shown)


app/presenters/caution_result_presenter.rb
```diff
@@ -3,6 +3,19 @@ def to_partial_path
     'results/caution'
   end
 
+  def variant
```
zheileman, line 6: No need to override this method. You only want the value of the `expiry_date` returned by the `CautionCalculator` to be `:simple` to accomplish the same.
Also, the `#caution_type` method to the bottom can be removed as will not be necessary.

app/value_objects/caution_type.rb
```diff
@@ -11,6 +11,13 @@ class CautionType < ValueObject
     ].freeze
   ].flatten.freeze
 
+  def simple?
```
zheileman, line 14: Because we've simplified (by removing it) the variant method, we don't need this method anymore.
Also, we are already producing the same result by negating the `#conditional?` method in some places.

Davidslv:
Closing because Github stopped updating this PR since I've added @njseeto as a co-author in the commits.
@zheileman I'll open another PR and I'm going to address your suggestions

PR #363, opened 2020-07-15
Simplify youth adult cautions journey

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/21186055

- Update caution decision tree
- Update caution result presenter

image

Co-authored-by: @njseeto


Davidslv:
I guess when one has a co-author, push force to amend commits is not allowed on GitHub

PR #366, opened 2020-07-28
Remove endorsement question from motoring penalty points

Davidslv:
Penalty points can only be accrued if the offence has been endorsed. So  you won’t see penalty points without endorsement.

Co-Authored-by: @njseeto 

Story: https://mojdigital.teamwork.com/#/tasks/21245933

### DIFF
--- a/app/forms/steps/conviction/conviction_subtype_form.rb
+++ b/app/forms/steps/conviction/conviction_subtype_form.rb
@@ -32,7 +32,8 @@ module Steps
           conviction_length: nil,
           conviction_length_type: nil,
           compensation_paid: nil,
-          compensation_payment_date: nil
+          compensation_payment_date: nil,
+          motoring_endorsement: nil
         )
       end
     end
--- a/app/services/conviction_decision_tree.rb
+++ b/app/services/conviction_decision_tree.rb
@@ -51,7 +51,9 @@ class ConvictionDecisionTree < BaseDecisionTree
   end
 
   def after_motoring_conviction
-    edit(:motoring_endorsement)
+    return edit(:motoring_endorsement) unless conviction_subtype.motoring_penalty_points?
+
+    known_date_question
   end
 
   def after_known_date
--- a/features/adults/conviction_motoring.feature
+++ b/features/adults/conviction_motoring.feature
@@ -40,8 +40,19 @@ Feature: Adult Conviction
       | Fine                       | 01-01-2020 | Yes         | When were you given the fine?            | will be spent on 1 January 2025 |
       | Fine                       | 01-01-2020 | No          | When were you given the fine?            | will be spent on 1 January 2021 |
       | Fixed Penalty notice (FPN) | 01-01-2020 | Yes         | When was the endorsement given?          | will be spent on 1 January 2025 |
-      | Penalty points             | 01-01-2020 | Yes         | When were you given the penalty points?  | will be spent on 1 January 2025 |
-      | Penalty points             | 01-01-2020 | No          | When were you given the penalty points?  | will be spent on 1 January 2023 |
+
+  @happy_path
+  Scenario Outline: Motoring convictions without length (penalty points)
+    When I choose "<subtype>"
+    Then I should see "<known_date_header>"
+
+    And I enter the following date <ban_date>
+    Then I should be on "/steps/check/results"
+    And I should see "This conviction <spent_date>"
+
+    Examples:
+      | subtype                    | ban_date   | known_date_header                        | spent_date     |
+      | Penalty points             | 01-01-2020 | When were you given the penalty points?  | will be spent on 1 January 2025 |
 
   @happy_path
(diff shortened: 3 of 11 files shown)


app/forms/steps/conviction/conviction_subtype_form.rb
```diff
@@ -19,12 +19,18 @@ def changed?
...
+        GenericYesNo::YES if ConvictionType.new(conviction_subtype).motoring_penalty_points?
+      end
+
       def persist!
         raise DisclosureCheckNotFound unless disclosure_check
         return true unless changed?
 
         disclosure_check.update(
           conviction_subtype: conviction_subtype,
+          motoring_endorsement: motoring_endorsement,
```
zheileman, line 33: It seems to me there is nothing in the service that requires this value to be set so, following the philosophy "no code, no bugs" I would recommend to remove this completely 😃 

The value will be set in the journey that asks the endorsement question, and any other journey will be `nil`.

Only difference is, in the results page, the line "Motoring endorsement: yes" will not show, but I think that's actually correct, as usually it is more confusing for the users to see information that they've not been asked for, specially if in this case is superfluous.

I would also add the `motoring_endorsement: nil` at the end of the list of dependent attributes to be reset, because if a user enters a motoring journey that asks this question but then go back and change to any other journey, the `motoring_endorsement` attribute might have been set and affect calculations or be confusing in the results page.

There might actually be more motoring attributes to be reset now that I see the list, but that come be done in a separate PR.

app/services/conviction_decision_tree.rb
```diff
@@ -45,7 +45,7 @@ def destination
   def after_conviction_subtype
     return edit(:conviction_bail)    if conviction_subtype.bailable_offense?
     return edit(:compensation_paid)  if conviction_subtype.compensation?
-    return after_motoring_conviction if conviction_type.motoring?
+    return after_motoring_conviction if conviction_type.motoring? && !conviction_subtype.motoring_penalty_points?
```
zheileman, line 48: I think it might be more clear if we leave the condition as it was, and add the different branching in the `after_motoring_conviction` method. This method is becoming a bit messy with so many conditions.

Like this:

```ruby
  def after_motoring_conviction
    return edit(:motoring_endorsement) unless conviction_subtype.motoring_penalty_points?

    known_date_question
  end
```

features/adults/conviction_motoring.feature
```diff
@@ -40,8 +40,19 @@ Feature: Adult Conviction
       | Fine                       | 01-01-2020 | Yes         | When were you given the fine?            | will be spent on 1 January 2025 |
       | Fine                       | 01-01-2020 | No          | When were you given the fine?            | will be spent on 1 January 2021 |
       | Fixed Penalty notice (FPN) | 01-01-2020 | Yes         | When was the endorsement given?          | will be spent on 1 January 2025 |
-      | Penalty points             | 01-01-2020 | Yes         | When were you given the penalty points?  | will be spent on 1 January 2025 |
-      | Penalty points             | 01-01-2020 | No          | When were you given the penalty points?  | will be spent on 1 January 2023 |
+
+  @happy_path
```
zheileman, line 44: Would be good to use the `@date_travel` and the `Given The current date is xxx` step here as well.

There are other cukes that will require it too but can be done as para of a separate PR as mentioned here: https://github.com/ministryofjustice/disclosure-checker/commit/184e8bb3c5f8b039a43b5e22744bdcd03d22d0f4

njseeto: Agreed. I have made a note and will address this in another PR.

app/services/conviction_decision_tree.rb
njseeto: Looks good. I've updated this file with this suggestion.

app/forms/steps/conviction/conviction_subtype_form.rb
njseeto: Yep, happy with "no code no bugs" philosophy!
I've made these changes.

I've also taken note on reseting the other motoring attributes and can address them in another PR.

spec/services/calculators/motoring/adult/penalty_points_spec.rb
```diff
@@ -6,20 +6,14 @@
   let(:disclosure_check) { build(:disclosure_check,
                                  under_age: under_age,
                                  known_date: known_date,
-                                 motoring_endorsement: motoring_endorsement) }
+                                 motoring_endorsement: GenericYesNo::YES) }
```
zheileman, line 9: As we don't use or set this attribute for penalty points anymore, probably best to cleanup this spec from any trace of endorsements to avoid confusion. Sorry I forgot to mention this in my first review!

Also, I realised the spec `motoring/youth/penalty_points_spec.rb` will require the same changes.

njseeto: Good spot! Have made the changes to this file and the spec `motoring/youth/penalty_points_spec.rb`

PR #368, opened 2020-07-30
Periodic maintenance of gems

zheileman:
Bump outdated gems, including rubocop, and fix a few new offenses.

Also, bumped govuk frontend to 3.8.0. It has minor changes, but some improve the accessibility.

Full list of changes here:
https://github.com/alphagov/govuk-frontend/releases/tag/v3.8.0


features/support/env.rb
```diff
@@ -5,6 +5,9 @@
 require 'cucumber/rails'
 require 'dotenv/load'
 
+# For some weird reason cucumber tests fail unless ActionMailer is required
```
njseeto, line 8: That is strange. 🤔 
I ran the cucumber tests both with and without ActionMailer and you're right everything fails without this.

zheileman: Yep! there must be something new introduced in cucumber 4.x (before we were using 3.x)
Not sure if it is a regression and they might fix it later on, will keep trying removing the `require` when bumping in the future cucumber, to see if it keeps failing or revert back to how was before 🤷‍♂️

PR #373, opened 2020-08-26
Design system upgrade

Davidslv:
We have:
-  installed the GOVUKDesignSystemFormBuilder gem.
- configured the error message to use GOVUKDesignSystemFormBuilder
- customised the govuk_submit button to use the existing YAML buttons namespace
- used the steps/check/kind edit page as a starting point

image

Story: https://mojdigital.teamwork.com/#/tasks/21650488


app/helpers/custom_form_helper.rb
```diff
@@ -0,0 +1,5 @@
+module CustomFormHelper
+  def submit_button(i18n_key, opts = {}, &block)
```
zheileman, line 2: Let's maintain if possible the same interface we had, because it will be needed when we get back to work on multiple convictions.

Basically, just replicating the `continue_button` (and `cancel_check_link`) methods we have here:
lib/govuk_components/form_builder.rb

But adapting them to use the new gem.

zheileman (changes requested):
Just a small comment. And probably we want tests for the form helper.

Other than that, great start! 🎉

zheileman:
This can go on another PR in the future, but we will also need to configure the new gem to output the tags and sizes we want, like we did on CAS.
For reference:
https://github.com/ministryofjustice/c100-application/blob/master/config/initializers/form_builder.rb

And the hint texts on the radios.

Hopefully with that it should look identical to this:

Screen Shot 2020-08-26 at 17 07 28

app/helpers/application_helper.rb
```diff
@@ -46,6 +46,17 @@ def error_summary(form_object = @form_object)
     )
   end
 
+  def govuk_error_summary(form_object = @form_object)
```
zheileman, line 49: Can go as well in another PR, but also realised the new helper is lacking the functionality to set the 'Error' prefix in the page title.

As per:
https://github.com/ministryofjustice/c100-application/blob/master/app/helpers/application_helper.rb#L26

Davidslv:
Update on the design, on the left side is localhost, right side is staging
image


We have updated the defaults of the DesignSystem so that there's less repetition around the forms (views).
I've removed the code related with `cancel_check_link` method because it does seem to be related with the multiple checks feature (which is not live and will need a revisit later on, so we can tackle that once we start working on that project).

Davidslv:
We have just realised we missing some proper testing around `govuk_error_summary`, we are going to look into it tomorrow.
We just added a WIP commit for the moment being

app/helpers/application_helper.rb
Davidslv: We will look into this again tomorrow, we realised it was missing after we realised there was not coverage on CI

lib/govuk_components/form_builder.rb
```diff
@@ -1,4 +1,9 @@
 # rubocop:disable Metrics/ClassLength
+# TODO: I've marked this module as not covered by tests because:
```
zheileman, line 2: Yep! good call.
Please do remove as much code as possible as you go along with the new gem. We had to implement many tests for all this custom code but going forward it will not be necessary as the new gem is heavily tested, we might just need a few cucumber fixture tests like we did on CAS but that can wait until the migration is more advanced 👍

app/helpers/application_helper.rb
zheileman: No worries! I think it may help with some failing tests, to use the same code we are using on CAS for this helper:
https://github.com/ministryofjustice/c100-application/blob/master/app/helpers/application_helper.rb#L26

Davidslv:
For reference, the changes necessary for the future pages are similar to the changes done in this commit https://github.com/ministryofjustice/disclosure-checker/pull/373/commits/0e03b3bb5b03c03a853ca9ab8d7cd06ecbd0fccf

zheileman (approved):
🚢 🚀

PR #374, opened 2020-08-28
Update layout to simplify markup 

zheileman:
Simplified each of the individual views with the absolute minimum necessary markup, lifting as much of it to the layout itself.

This also place the back link right where it needs to go, above the `main` container.

This looks like a huge PR but in reality is just removing a couple containers and changing the indentation of the markup 🤣 

These 2 outer containers are removed in each view:
Screen Shot 2020-08-28 at 13 25 53


njseeto (approved):
nice, big clean up! ✨

app/views/layouts/application.html.erb
```diff
@@ -1,6 +1,7 @@
 <% content_for?(:page_title) ? yield(:page_title) : fallback_title %>
 
 <% content_for(:head) do %>
+  <%= csrf_meta_tags %>
```
Davidslv, line 4: 👍

PR #375, opened 2020-08-28
Migrate caution dates to use new gem

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21679176

This maintains compatibility with the old dates, so both can coexist.
Some tests duplications until we've migrated all dates but that is ok.

**Note: this uses code created and tested in the previous CAS migration.**

As caution only uses dates in one journey (conditional caution), this finalises the dates migration for cautions, there are no more.

Convictions will be next...

### Before
Screen Shot 2020-08-28 at 16 20 23

### After
Screen Shot 2020-08-28 at 16 20 01


njseeto (approved):
nice one! ✨

PR #376, opened 2020-09-01
Migrate caution type to use new gem

njseeto:
Task: https://mojdigital.teamwork.com/#/tasks/21707821

### Before (adult)
Screenshot 2020-09-01 at 16 55 19

### After (adult)
after adult

### Before (youth)
before youth

### After (youth)
after youth

### Before (youth error)
Screenshot 2020-09-02 at 10 35 23

### After (youth error)
youth new

### Before (adult error)
adult old

### After (adult error)
adult new

### DIFF
--- a/app/views/steps/caution/caution_type/edit.html.erb
+++ b/app/views/steps/caution/caution_type/edit.html.erb
@@ -1,21 +1,14 @@
 <% title t('.page_title') %>
+<% step_header %>
 
-<div class="govuk-width-container">
-  <%= step_header %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary %>
 
-  <main class="govuk-main-wrapper" id="main-content">
+    <%= step_form @form_object, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+      <%= f.govuk_collection_radio_buttons :caution_type, @form_object.values, :value, nil %>
 
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
-        <%= error_summary %>
-
-        <%= step_form @form_object do |f| %>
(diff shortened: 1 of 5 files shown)


Davidslv:
@njseeto can you please also take a screenshot of the error message?

Thank you!

config/locales/en/helpers.yml
```diff
@@ -25,6 +25,12 @@ en:
       adult_simple_caution: Simple caution
       adult_conditional_caution: Conditional caution
 
+    CAUTION_TYPE_HINTS: &CAUTION_TYPE_HINTS
```
zheileman, line 28: Maybe the dictionary can have just 2 keys, one for `simple_caution_hint` and another for `conditional_caution_hint` and then we use these values in the `caution_type_options`?
Just an idea. Otherwise a dictionary seems a bit like of no use.
Also we can delete the old hint texts as they are no longer used I believe.

zheileman (approved):
LGTM 🥇 🎉 
Just a couple suggestions, not important.

app/views/steps/caution/caution_type/edit.html.erb
```diff
@@ -1,21 +1,14 @@
...
 
-<div class="govuk-width-container">
-  <%= step_header %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary %>
 
-  <main class="govuk-main-wrapper" id="main-content">
+    <%= step_form @form_object, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+      <%= f.govuk_collection_radio_buttons :caution_type, @form_object.values, :value, nil %>
```
Davidslv, line 9: add the following to the end: `, inline: true`
So that we keep the radio buttons as they were before, please :)

config/locales/en/helpers.yml
njseeto: Ah yes, good shout, will do this!

PR #377, opened 2020-09-02
Update Design System for steps/conviction/compensation_paid_amount

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/21707829

### Before
Screenshot 2020-09-02 at 08 54 49

### After

Screenshot 2020-09-02 at 08 54 31

### DIFF
--- a/app/views/steps/conviction/compensation_paid_amount/edit.html.erb
+++ b/app/views/steps/conviction/compensation_paid_amount/edit.html.erb
@@ -1,23 +1,15 @@
 <% title t('.page_title') %>
+<% step_header %>
 
-<div class="govuk-width-container">
-  <%= step_header %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary %>
+    <%= step_subsection %>
 
-  <main id="main-content" class="govuk-main-wrapper">
+    <%= step_form @form_object, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+      <%= f.govuk_collection_radio_buttons :compensation_payment_over_100, GenericYesNo.values, :value, nil, inline: true %>
 
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
-        <%= error_summary %>
-        <%= step_subsection %>
-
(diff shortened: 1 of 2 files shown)


app/views/steps/conviction/compensation_paid_amount/edit.html.erb
```diff
@@ -1,23 +1,15 @@
 <% title t('.page_title') %>
+<%= step_header %>
```
zheileman, line 2: Super minor thing, it doesn't really affect in any visible way but for the `step_header` we can now use the same syntax as with the `title` (removing the `=`) as it does not output anything in place, it just assigns the content for the `back_link`

app/views/steps/conviction/compensation_paid_amount/edit.html.erb
```diff
@@ -1,23 +1,15 @@
...
 
-<div class="govuk-width-container">
-  <%= step_header %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary %>
+    <%= step_subsection %>
```
zheileman, line 7: I think for now it is ok to continue using this helper, until we've finished all the steps. Then we can go back and change this to use the native `caption` that the gem offers.

zheileman (approved):
👍 💯 
Couple of minor comments but not affecting the PR.

app/views/steps/conviction/compensation_paid_amount/edit.html.erb
Davidslv: Updated on https://github.com/ministryofjustice/disclosure-checker/pull/377/commits/3e9fc6b34d11c19de05b11512015dd310228b2d3

app/views/steps/conviction/compensation_paid_amount/edit.html.erb
Davidslv: addressed

app/views/steps/conviction/compensation_paid_amount/edit.html.erb
Davidslv: Got a story for this one: https://mojdigital.teamwork.com/#/tasks/21724564

njseeto (approved):
nice! ✨

PR #378, opened 2020-09-02
Update Design System for steps/conviction/compensation_paid

Davidslv:
Needs to be rebased after https://github.com/ministryofjustice/disclosure-checker/pull/377 is merged to use the YAML constant YESNO.

Story: https://mojdigital.teamwork.com/#/tasks/21707828

### Before
Screenshot 2020-09-02 at 09 07 46

### After

Screenshot 2020-09-02 at 09 07 22

### DIFF
--- a/app/views/steps/conviction/compensation_paid/edit.html.erb
+++ b/app/views/steps/conviction/compensation_paid/edit.html.erb
@@ -1,23 +1,15 @@
 <% title t('.page_title') %>
+<% step_header %>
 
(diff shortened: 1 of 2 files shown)


zheileman (approved):
💯 🎉 

I'm thinking, because in this service we don't have revealing radios, all these Yes-No questions perhaps can use inline radios? (what we had before).
Maybe worth checking with Simon 👍

njseeto (approved):
LGTM! 🚀

PR #380, opened 2020-09-02
Migrate `conviction_known_date` to use new gem

zheileman:
Similar to previous PR to migrate gems, but this one is a bit special as this is the first form to be migrated with our custom `i18n_attribute` that we created to allow a same form to be reused in multiple views with different legends and hint texts.

This is a preliminary version of a potential solution that more or less maintain the same interface and same locales and should be easy to migrate the rest of the forms.

Screen Shot 2020-09-02 at 16 10 34

Screen Shot 2020-09-02 at 16 10 45


Davidslv (approved):
Very nice, thank you!

PR #381, opened 2020-09-02
Update Design System for steps/conviction/compensation_payment_receipt

Davidslv:
Screenshot 2020-09-02 at 16 10 46

Story: https://mojdigital.teamwork.com/#/tasks/21707831


zheileman (approved):
👍 💯

config/locales/en/helpers.yml
```diff
@@ -161,6 +161,12 @@ en:
         adult_disqualification: Was the length of the disqualification given in weeks, months or years?
       steps_conviction_compensation_payment_receipt_form:
         compensation_receipt_sent: Did you send the payment receipt to the Disclosure and Barring (DBS) service?
+      steps_conviction_compensation_paid_form:
```
zheileman, line 164: I think some of these are dupes. I keep removing them in my PRs but they... reappear? 🤣

Davidslv: I think it may have been during my rebase, let's catch up on this together so I remove the right stuff 👍 
Well spotted!

Davidslv:
Much cleaner now 😄 without those duplicated locales

PR #384, opened 2020-09-03
Update Design System for under_age

njseeto:
Task: https://mojdigital.teamwork.com/#/tasks/21707824

### Before
old check_age

### After
new age check

### Before
old error

### After
new error

### Before
old

### After
new

### Before
old

### After
Screenshot 2020-09-03 at 15 46 03


zheileman (approved):
💯 🚢 Great!

PR #385, opened 2020-09-03
Update Design System for steps/conviction/conviction_subtype

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/21725158

Screenshot 2020-09-03 at 15 48 27

Screenshot 2020-09-03 at 15 48 19

### DIFF
--- a/app/views/steps/conviction/conviction_subtype/edit.html.erb
+++ b/app/views/steps/conviction/conviction_subtype/edit.html.erb
@@ -1,23 +1,15 @@
 <% title t('.page_title') %>
+<% step_header %>
 
-<div class="govuk-width-container">
-  <%= step_header %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary %>
 
-  <main id="main-content" class="govuk-main-wrapper">
+    <%= step_form @form_object, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+      <%= f.govuk_collection_radio_buttons :conviction_subtype, @form_object.values, :value, nil,
+            legend: { text: f.i18n_legend } %>
 
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
-        <%= error_summary %>
-
-        <%= step_form @form_object do |f| %>
-          <%= f.radio_button_fieldset :conviction_subtype, radio_hint: true,
-                                      choices: @form_object.choices,
-                                      i18n_attribute: @form_object.conviction_type %>
-
-          <%= f.continue_button %>
-        <% end %>
-      </div>
-    </div>
-
-  </main>
+      <%= f.continue_button %>
+    <% end %>
+  </div>
 </div>
--- a/config/locales/en/helpers.yml
+++ b/config/locales/en/helpers.yml
@@ -128,21 +128,6 @@ en:
       cancel_check: Go back to check your answers
       results_page: Go to results page
     fieldset:
-      steps_conviction_conviction_subtype_form:
-        # default key
-        conviction_subtype: What type of order were you given?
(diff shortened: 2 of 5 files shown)


app/views/steps/conviction/conviction_subtype/edit.html.erb
```diff
@@ -1,23 +1,15 @@
...
-<div class="govuk-width-container">
-  <%= step_header %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary %>
 
-  <main id="main-content" class="govuk-main-wrapper">
+    <%= step_form @form_object, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+      <%= f.govuk_collection_radio_buttons :conviction_subtype, @form_object.values, :value, nil,
+            legend: { text: f.i18n_legend } %>
```
zheileman, line 10: This requires also hints (anything that before had `radio_hint: true`) using the `i18n_hint`
Otherwise hints will not show.

config/locales/en/helpers.yml
```diff
@@ -382,6 +372,22 @@ en:
       steps_conviction_motoring_disqualification_end_date_form:
         motoring_disqualification_end_date: When did your disqualification end?
         approximate_motoring_disqualification_end_date: *approximate_date_legend
+      steps_conviction_conviction_subtype_form:
+        # default key
+        adult_financial: What type of order were you given?
```
zheileman, line 377: `adult_financial` and `prevention_reparation` should be replaced with:

```yml
default: What type of order were you given?
```

So the legend is shown if no specific key is defined.

config/locales/en/helpers.yml
```diff
@@ -128,21 +128,11 @@ en:
...
-        discharge: What discharge were you given?
-        financial: What were you ordered to pay?
-        military: What was your military conviction?
-        adult_military: What was your military conviction?
-        adult_motoring: What was your motoring conviction?
-        youth_motoring: What was your motoring conviction?
-        adult_community_reparation: What was your community, prevention or reparation order?
-        adult_discharge: What discharge were you given?
-        adult_custodial_sentence: What sentence were you given?
+      steps_check_under_age_form:
```
zheileman, line 131: `steps_check_under_age_form` and `steps_conviction_conviction_type_form` were again reintroduced after a bad merge. Need to be removed, these are already present in the `legend` section.

zheileman (changes requested):
A few changes needed 👍

config/locales/en/helpers.yml
```diff
@@ -223,8 +213,8 @@ en:
         conviction_bail: Time spent on bail with an electronic tag affects the spent date of your conviction. Time spent on bail without a tag does not.
       steps_conviction_conviction_bail_days_form:
         conviction_bail_days: This is decided by the court. If you do not know the exact number of days you can enter an approximate number or leave this blank, but it might affect the accuracy of your result.
-      radio_buttons:
-        conviction_subtype:
+      steps_conviction_conviction_subtype_form:
+        conviction_subtype_options:
```
zheileman, line 217: Before it was not possible but I think now with the new gem is possible to just remove the hint key when we don't want to show a hint, instead of having an empty string `fine: ""` can you try? That way we clean up a lot of those empty keys in this group.

app/views/steps/conviction/conviction_subtype/edit.html.erb
Davidslv: No worries, I was confused as well when I did it

config/locales/en/helpers.yml
Davidslv: It was not catching the default, I probably need to change some code, I'll have a look now

zheileman (approved):
I think all is good now 💯

PR #387, opened 2020-09-04
Upgrade Design System for steps/conviction/conviction_bail

njseeto:
Task: https://mojdigital.teamwork.com/#/tasks/21725161

### Before
old bail

### After
new bail

### Before
old error

### After
new error


zheileman (approved):
💥

PR #390, opened 2020-09-04
Update Design System for steps/conviction/motoring_endorsement

njseeto:
Task: https://mojdigital.teamwork.com/#/tasks/21742527

#### Before
old endorsement

#### After
new endorsement

#### Before
old error

#### After
Screenshot 2020-09-04 at 12 49 31

### DIFF
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -65,6 +65,10 @@ en:
               greater_than: The input must be greater than zero, or left blank
               not_a_number: The input must be a number, like 3
               not_an_integer: The input must be a whole number, like 4
+        steps/conviction/motoring_endorsement_form:
+          attributes:
+            motoring_endorsement:
+              inclusion: Select yes if you have received an endorsement
 
   errors:
     format: "%{message}"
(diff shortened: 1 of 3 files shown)


zheileman (changes requested):
It looks like the error message was missing from before, let's fix this in this PR, is an easy thing to add to the locales.

njseeto:
Good spot @zheileman! I have added an error message in the new version, am happy to change the wording if we think it needs to be adjusted.

config/locales/en/errors.yml
```diff
@@ -65,6 +65,10 @@ en:
               greater_than: The input must be greater than zero, or left blank
               not_a_number: The input must be a number, like 3
               not_an_integer: The input must be a whole number, like 4
+        steps/conviction/motoring_endorsement_form:
+          attributes:
+            motoring_endorsement:
+              inclusion: Select yes if you have received an endorsement.
```
zheileman, line 71: Yep perfect! Maybe we can remove the full stop as it seems we don't use them for error messages anywhere, but that's a minor thing!

zheileman (approved):
👍 🏅

PR #391, opened 2020-09-04
Update Design System for steps/conviction/conviction_bail_days/

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/21725162

There's a bit of a discrepancy between before and now:
- 'number of days' is not bold (I've tried to play with the size option but nothing changed, ideas?)
- title of the page used to be larger
- On C100 we normally used `one-half` or similar, but any of those sizes made the text field look really wide, so I used `width: 3` instead

Left is new design, right is old design

Screenshot 2020-09-04 at 15 21 29


zheileman (approved):
💯 🥇

PR #392, opened 2020-09-04
Update Design System for steps/conviction/conviction_length

njseeto:
Task: https://mojdigital.teamwork.com/#/tasks/21742522

#### Before
old

#### After
new

#### Before
old error

#### After
new error


config/locales/en/helpers.yml
```diff
@@ -272,9 +272,9 @@ en:
           years: Years
           no_length: No length was given
       steps_conviction_conviction_length_form:
-        weeks: Number of weeks
-        months: Number of months
-        years: Number of years
+          weeks: Number of weeks
```
Davidslv, line 275: There seems to be one level deeper than necessary (one less tab here)

njseeto: good spot!

Davidslv (approved):
LGTM 👍

PR #393, opened 2020-09-07
Define GOVUKDesignSystemFormBuilder::FormBuilder as the default

Davidslv:
This commit removes all the explicit places where we had previously defined which form builder to be used and make it the default form builder in this project.


njseeto (approved):
Just one small thing, do we need to take a look at this file too `app/helpers/application_helper.rb`?

There is a TODO in there:
```
    # TODO: can be removed once we use this builder globally
    options = {
      builder: GOVUKDesignSystemFormBuilder::FormBuilder
    }
```

Other than that looks good! ✨

zheileman (approved):
👍

PR #394, opened 2020-09-08
Use the caption field provided by the Design System gem

Davidslv:
The caption is now part of the form, that means that if there's an error message the whole vertical red line also covers the caption above the page header (h1)

Example:

Screenshot 2020-09-08 at 10 51 15

Story: https://mojdigital.teamwork.com/#/tasks/21724564

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ tags
 
 public/assets/
 .DS_Store
+
+rerun.txt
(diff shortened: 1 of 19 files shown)


zheileman (approved):
👍 🎉 LGTM!

Davidslv:
Screenshot of the text field with fieldset

Screenshot 2020-09-08 at 14 02 59

PR #395, opened 2020-09-09
add missing translation key for compensation paid page

Davidslv:
Missed the compensation paid page caption translation key.
This did not affect production (just staging)


njseeto (approved):
💯 💯 💯

PR #396, opened 2020-09-09
Remove govuk components gem and cleanup

Davidslv:
I've removed the govuk_components gem and also noticed a few places where we should be using the `govuk_error_summary` 

Story: https://mojdigital.teamwork.com/#/tasks/21793490


zheileman (approved):
💯 soooo clean now! 🤣

njseeto (approved):
so fresh and so clean! ✨

PR #397, opened 2020-09-10
Update GA dates tracking to work with new gem

zheileman:
The new gem produce a slighly different markup with multiparam text inputs.

In order to keep tracking these (only the year part) as before, with the same event category, we need to do some tweaks.


njseeto (approved):
LGTM!

PR #400, opened 2020-10-05
Update design system to version 2.1.0

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/21794134

DFE Update documentation: https://github.com/DFE-Digital/govuk_design_system_formbuilder/releases/tag/v2.0.0

The only bit I'm not understanding what I should do is the [legend wrapped around the h1](https://github.com/DFE-Digital/govuk_design_system_formbuilder/releases/tag/v2.1.0) since we currently have the following configuration:

```ruby

GOVUKDesignSystemFormBuilder.configure do |config|
  config.default_legend_tag   = 'h1' # <<<< here
  config.default_legend_size  = 'xl'
  config.default_caption_size = 'xl'
end
```


app/views/steps/shared/_approx_date_checkbox.html.erb
```diff
@@ -1,3 +1,3 @@
 <%= form.govuk_check_boxes_fieldset attribute, small: true, legend: { tag: 'span', size: 's', hidden: true } do %>
-  <%= form.govuk_check_box attribute, true, multiple: false, link_errors: true %>
+  <%= form.govuk_check_box attribute, true, false, multiple: false, link_errors: true %>
```
Davidslv, line 2: This ensures forms with #govuk_check_boxes_fieldset are able to be submitted with a present but empty param instead of none. There is now a valueless #hidden_field added to the form to represent an unchecked state

zheileman:
@Davidslv Let me checkout the branch and take a look but I think our default for legends will be fine and should keep working as it is now 👍

app/views/steps/shared/_approx_date_checkbox.html.erb
```diff
@@ -1,3 +1,3 @@
 <%= form.govuk_check_boxes_fieldset attribute, small: true, legend: { tag: 'span', size: 's', hidden: true } do %>
```
zheileman, line 1: Just a few changes to fully be compatible with the new version.

We need to add the `multiple: false` to the `govuk_check_boxes_fieldset`
as per documentation, otherwise the hidden field will have an incorrect name:

https://govuk-form-builder.netlify.app/form-elements/checkboxes/#generating-a-single-checkbox

And we also now can `nil` the legend tag so it does not generate an `h1` (nor a `span` like before).

```suggestion
<%= form.govuk_check_boxes_fieldset attribute, multiple: false, small: true, legend: { tag: nil, size: 's', hidden: true } do %>
```

zheileman (changes requested):
Great! 🎉 
Not a very painful upgrade thankfully 👍 
I just added a comment to tweak the check box to fully be inline with the new gem version.

zheileman (approved):

PR #401, opened 2020-10-07
Update rails version to 5.2.4.4

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/22085126


Gemfile
```diff
@@ -14,7 +14,7 @@ gem 'govuk_design_system_formbuilder'
 gem 'jquery-rails'
 gem 'pg', '~> 1.1'
 gem 'puma'
-gem 'rails', '~> 5.2.4'
+gem 'rails', '~> 5.2.4.4'
```
zheileman, line 17: Doing this will restrict future bumps as this is reducing minor releases to the last .x version (so for example, if tomorrow version 5.2.5 is released, this will forbid the update despite still being considered a minor release).

We want any version in the 5.2 branch thus the `~> 5.2.4` so it considers any minor released in the 5.2.x.x range.

So no need to change this really, just doing a `bundle update rails` will figure it out automatically which version it can be updated and apply the changes to the lock file.

zheileman (approved):
👍 
Just a comment regarding the versioning.

Also, if you want to try, there are lots of other bumps to other gems, doing a `bundle update` you will notice. But some require a few changes or locking versions if we are yet not ready to update.
It is necessary to read through the CHANGELOG of each major version to see the changes and how it impacts us.

Davidslv:
I'll update this one and then look into the other updates if there's anything that breaks or something

PR #402, opened 2020-10-08
Upgrade cucumber rails and other gems

Davidslv:
This is a follow up on #401 

List of circle ci images here: https://circleci.com/docs/2.0/circleci-images/#ruby
Dockerfile ruby images available here: https://hub.docker.com/_/ruby


zheileman:
Dockerfile will also need an update to use the new ruby version
currently is using `FROM ruby:2.6.5-alpine`
as well as bundler `bundler -v 1.17.3` etc.

Gemfile.lock
```diff
@@ -76,47 +77,32 @@ GEM
       descendants_tracker (~> 0.0.1)
     concurrent-ruby (1.1.7)
     crass (1.0.6)
-    cucumber (4.1.0)
```
Davidslv, line 79: Computer said no to upgrading cucumber?
Investigating this

zheileman (approved):
🎉 great!
Now I hope I don't break my machine like the other day trying to get ruby 2.7.2 working 🤣

Gemfile.lock
Davidslv: Ok, for some reason I though I was on cucumber-rails 1.4.1, now looking at the commit history I can see that I did not upgrade the project to 1.6.0 neither 1.8.0 as Gemfile.lock had 2.1.0 version (I'm not sure where I took this idea from), that then caused cucumber to be downgraded (obviously)

PR #403, opened 2020-10-08
Add .ruby-version file in Dockerfile

Davidslv:
Running the docker image won't know about the `.ruby-version` file without copying it into the image.
This happened on #402 when I changed the Gemfile to read the ruby version from that file.

I should have run `docker-compose up` to see the problem.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,7 @@ RUN addgroup -g 1000 -S appgroup && \
 RUN mkdir -p /usr/src/app && mkdir -p /usr/src/app/tmp
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -35,6 +35,7 @@ RUN addgroup -g 1000 -S appgroup && \
 RUN mkdir -p /usr/src/app && mkdir -p /usr/src/app/tmp
 WORKDIR /usr/src/app
 
+COPY Gemfile* .ruby-version ./
 COPY Gemfile* ./
```
zheileman, line 39: Sorry I missed this.
Now there is a duplication. This has to be removed although does not break the image.

PR #407, opened 2020-11-05
Bump design system version

zheileman:
As well as updating to the latest version of `govuk-frontend`.
Some other gem minor bumps.

Fixed a deprecation warning in bundler.

Made a couple yes/no radios inline, consistent with the rest of pages where we also had yes/no as inline.


Davidslv (approved):
👍

PR #408, opened 2020-11-11
Design System smoke tests

Davidslv:
During an upgrade to the Design System on C100 it was noticed that the forms were not working properly with hints on check boxes (https://github.com/DFE-Digital/govuk_design_system_formbuilder/pull/211).

We decided to introduce the same mechanism used on C100 in Disclosure Checker.

Story: https://mojdigital.teamwork.com/#/tasks/22441209

### DIFF
--- a/features/step_definitions/common.rb
+++ b/features/step_definitions/common.rb
@@ -50,8 +50,7 @@ And(/^I choose "([^"]*)" and fill in "([^"]*)" with "([^"]*)"$/) do |text, field
 end
 
 When(/^I am completing a basic under 18 "([^"]*)" conviction$/) do |value|
-  step %[I visit "/"]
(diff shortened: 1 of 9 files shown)


features/step_definitions/common.rb
```diff
@@ -61,8 +60,7 @@
 end
 
 When(/^I am completing a basic 18 or over "([^"]*)" conviction$/) do |value|
-  step %[I visit "/"]
-  step %[I click the "Start now" link]
+  step %[I have started an application]
```
zheileman, line 63: Super minor thing, but maybe instead of `an application` this might be a `a check`, for this service?

zheileman (approved):
👍 💯 
At least we have a little safety net for the future.

PR #414, opened 2021-01-07
Identify which sub conviction types are relevant orders

Davidslv:
This change won't affect any existing calculations or journeys. This is a pre step purely to identify which sub conviction types are considered relevant orders. This will be important to help calculations when a user is given a conviction with several sentences, if one of them is a relevant order, that sentence is not considered - meaning that even if ends later than all others, it won't affect the spent end date.

**Please do note** that the codebase defines a Sentence as a Conviction, although in real life it works as such:

A person is given a Conviction and this may have 1 or more sentences within.

Story: https://mojdigital.teamwork.com/#/tasks/23009279
Supporting Document: https://docs.google.com/spreadsheets/d/1ZSCk-wgMfIc22GQahKH_ys4u-E9GWIn6dQ7mm5t1RFI/edit#gid=2019860123

image

### DIFF
--- a/app/value_objects/conviction_type.rb
+++ b/app/value_objects/conviction_type.rb
@@ -7,6 +7,7 @@ class ConvictionType < ValueObject
     @parent = params.fetch(:parent, nil)
     @skip_length = params.fetch(:skip_length, false)
     @calculator_class = params.fetch(:calculator_class, nil)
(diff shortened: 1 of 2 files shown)


zheileman (approved):
💯

PR #417, opened 2021-01-15
Fix BaseMultiplesCalculator to consider spent_simple

Davidslv:
Back in pull request [#364](https://github.com/ministryofjustice/disclosure-checker/pull/364) the simple caution was simplified and the date question was skipped.

While this is currently working, this work is necessary in preparation for multiple convictions.

This fixes the following error:

```
NoMethodError:
       undefined method `past?' for :spent_simple:Symbol
```

When a user has a simple caution, that caution is spent on the day it was given, therefore the return value is true.

Story: https://trello.com/c/Gbnpo8Hu

### DIFF
--- a/spec/services/calculators/multiples/multiple_offenses_calculator_spec.rb
+++ b/spec/services/calculators/multiples/multiple_offenses_calculator_spec.rb
@@ -56,6 +56,14 @@ RSpec.describe Calculators::Multiples::MultipleOffensesCalculator do
       end
     end
 
+    context 'when there is an offence with `spent_simple`' do
+      let(:spent_dates) { [Date.yesterday, :spent_simple] }
+
+      it 'considers the spent_simple as spent' do
+        expect(subject.all_spent?).to eq(true)
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


Davidslv:
[[DC] Investigate how simple cautions may break calculations for multiple convictions](https://trello.com/c/Gbnpo8Hu/56-dc-investigate-how-simple-cautions-may-break-calculations-for-multiple-convictions)

spec/services/calculators/multiples/multiple_offenses_calculator_spec.rb
```diff
@@ -56,6 +56,14 @@
       end
     end
 
+    context 'when there is an offence with `spent_simple`' do
+      let(:spent_dates) { [:spent_simple, :spent_simple] }
```
zheileman, line 60: Would be better to add a date to this array. The test in theory should behave identical to line 70, if there is one (or more) `spent_simple` those are considered spent, but then we use any other date in the array.

zheileman (approved):
👍 just a minor suggestion!

PR #418, opened 2021-01-18
Add an "indefinite" length option to relevant orders

zheileman:
Ticket: https://trello.com/c/MrEm9UKp

Individual commits for more detail.

Relevant orders need to expose a new "length" option apart from the existing ones. In total now there are up to 5 different options: weeks, months, years, no length, indefinite (until further order).

If the conviction has an indefinite length, then a different results page, specific for indefinite is shown to the user.

**Note**: 18 out of 20 relevant orders now have the ability of selecting indefinite option. We still have 2 more that requires it but will be done as part of a separate ticket as will require design and probably a new question to ask. These are motoring disqualifications (youths and adults).

Screenshot 2021-01-18 at 12 27 02

### DIFF
--- a/app/presenters/results_presenter.rb
+++ b/app/presenters/results_presenter.rb
@@ -42,7 +42,8 @@ class ResultsPresenter
               expiry_date
             end
 
-    # The tense can be one of these values: spent, not_spent, never_spent or no_record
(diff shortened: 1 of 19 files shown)


Davidslv (approved):
LGTM 👍

PR #419, opened 2021-01-18
Remove motoring Fixed Penalty Notice

Davidslv:
A decision has been made to remove FPN from the service as these notices are spent on the day that they are given. 
This commit removes all the references to either Youth or Adult FPN convictions.

All done in one commit to be easier to revert (if ever needed).

Story: https://trello.com/c/Vcpqhlqr

PR #420, opened 2021-01-19
Delete some old code no longer needed for `no_record`

zheileman:
We've removed FPN in PR #419, so we don't use anymore the special result `no_record`.
Some code can be cleaned up.

As part of a following PR I'm refactoring the different variants we have (never_spent, spent_simple, indefinite, etc) as it is becoming increasingly difficult to find all the places where we are using them and will be easier to work with a single value-object.


Davidslv (approved):
👍 💯

PR #421, opened 2021-01-19
Refactor and DRY the different result variants

zheileman:
Ticket: https://trello.com/c/LS4gr8Uh

Over the last few months, we've had to add some "exceptions" to the rule and some cautions or convictions do not return just a date when they are spent, some returns special variants, like:

* never_spent
* spent_simple
* indefinite

We also had before `no_record` for FPN but this is no longer necessary as we've removed FPN from the service.

Unify in the `ResultsVariant` value-object all these possible result variants and always use them through the value-object so it is easy to either remove them, add more or change them in the future.

### DIFF
--- a/app/presenters/results_presenter.rb
+++ b/app/presenters/results_presenter.rb
@@ -37,13 +37,12 @@ class ResultsPresenter
 
   def variant
     tense = if expiry_date.instance_of?(Date)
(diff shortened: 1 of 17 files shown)


Davidslv (approved):
👍

PR #422, opened 2021-01-20
Refactor motoring disqualification calculator

zheileman:

This is some preliminary work to simplify the implementation of the `indefinite` length option, and also to reduce the repetition of code.

There are no changes to the rules or the implementation of the calculators, just a refactor to make them more reusable (and look more similar to other calculators like the SentenceCalculator).

In a follow-up PR, we will change the decision tree for motoring disqualifications to ask for the ban length (instead of the ban end date) and update the calculators to work with the length.


Davidslv (approved):
👍

PR #423, opened 2021-01-21
Implement "indefinite" length for motoring disqualifications

zheileman:

This is a follow-up to PR #351.
Individual commits for more details.

For motoring disqualifications we should also support an "indefinite / until further order" option, but because motoring disqualifications did not "flow" through the conviction length type step as other orders, they did not get the "indefinite" functionality we implemented in PR #418.

With this PR we change the disqualification journey to ask for a length, instead of an end date, and update the calculator to work with this. This way we also expose the `no length was given` and the `until further order` options.

**Note**: in a future PR we will remove the unused step/views/locales, and the DB attribute.

Screenshot 2021-01-21 at 12 12 15

### DIFF
--- a/app/decorators/conviction_decorator.rb
+++ b/app/decorators/conviction_decorator.rb
@@ -18,11 +18,6 @@ module ConvictionDecorator
       ConvictionType::ADULT_MOTORING.eql?(self)
   end
 
-  def motoring_disqualification?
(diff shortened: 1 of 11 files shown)


Davidslv (commented):
Awesome! 👍 
Thank you!

Davidslv (approved):
aprove! 😄

PR #424, opened 2021-01-22
Cleanup unused motoring disqualification end date

zheileman:

We replaced this step with the length of the disqualification in PR #423, so the "no length was given" and "until further order" are also options available.

Cleanup unused code and DB attribute.


Davidslv (approved):
Approve 💯

PR #427, opened 2021-01-26
First attempt to perform drag-through calculations

zheileman:

This is a first (and admittedly not very elegant) attempt to perform drag-through calculations for **separate proceedings**, meaning we need to extend the spent date of a conviction if the rehabilitation period of said conviction overlaps (even if it is just 1 day) with the rehabilitation period of another conviction.

Note, some limitations with this first version that will be sorted out in follow-up PRs:

- The loop is not sorted by conviction date, so the final result is not idempotent (depends on the order the convictions are added to the "basket"). This loop will need to be sorted.

- We will need to do something similar, if not identical, in the `SameProceedings` calculator, because the sentences inside a conviction can also have drag-through, but there are some exceptions.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -52,6 +52,12 @@ Metrics/MethodLength:
 Metrics/AbcSize:
   Max: 20
 
+Metrics/CyclomaticComplexity:
+  Max: 9
+
+Metrics/PerceivedComplexity:
+  Max: 9
+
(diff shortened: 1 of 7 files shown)


Davidslv (approved):
I don't have a better solution at the moment.
I think it's better to have something that works right and then we can eventually improve it.

👍

PR #428, opened 2021-01-26
Sort drag-through loop by conviction start date

zheileman:

Follow-up to PR #427.

As mentioned before the results were not idempotent. This is now solved by having a predictable iteration order in the loop, by conviction `start date`.

The start date for a conviction with more than one sentences is the earliest of the dates.

Some of the scenarios in the Unlock spreadsheet can now be coded into tests with factories and should give us predictable (and accurate) results.

Other scenarios can't be run just yet as they will fail.


Davidslv (approved):
👍

PR #429, opened 2021-01-27
Multiple offences calculator tweaks

zheileman:

Follow-up to PR #428.

Using a range is more accurate as there were some edge cases where otherwises the loop would return wrong results.

Also, tweak the multiples journey to not ask again for the conviction date, in same proceedings, as the start date is shared between all sentences.


Davidslv (approved):
👍

PR #430, opened 2021-01-27
Create specs based on given examples

Davidslv:
This commit brings a new set of tests that have been based on some scenarios,
so that we can improve the calculatios of multiple convictions (same proceedings or separte proceedings)

I've marked a test as pending for same proceedings since we haven't tackled yet relevant orders.

Although another test showed that we are marking all separate convictions as
never spent if there's a never spent conviction in one of the groups.

This is not correct, only the convictions that overlap would be considered never spent,
any convictions which are spent before should not be considered "never spent".

It may be better to bring the screenshot explaining the scenario with 4 convictions

image

**This Pull Request will fail** but I'd like to confirm this before attempting a fix on the calculator.

Story: https://trello.com/c/W2K1PV43/40-dc-calculations-for-separate-proceedings

### DIFF
--- a/app/models/disclosure_check.rb
+++ b/app/models/disclosure_check.rb
@@ -6,4 +6,8 @@ class DisclosureCheck < ApplicationRecord
     in_progress: 0,
     completed: 10,
   }
+
+  def relevant_order?
+    ConvictionType.find_constant(conviction_subtype).relevant_order?
+  end
 end
--- a/app/services/calculators/multiples/same_proceedings.rb
+++ b/app/services/calculators/multiples/same_proceedings.rb
@@ -17,10 +17,41 @@ module Calculators
 
       private
 
+      # TODO: This needs confirmation, please visit the following card https://trello.com/c/sZ7qBDwe/
       def expiry_dates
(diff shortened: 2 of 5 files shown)


zheileman:
Hey yep, I'm aware of the never spent / indefinite being overzealous and taking over other spent days 🤣 
As part of another PR I've done a bit of refactor and sorting and should be now be easier to implement this special handling hopefully in some future PR.

With the refactor some of the multiples calculator interface has changed so instead of passing a `check_group` you now pass the proceeding itself. And no need to call `#process!` either.

Davidslv:
[[DC] Calculations for Separate Proceedings](https://trello.com/c/W2K1PV43/40-dc-calculations-for-separate-proceedings)

Davidslv:
[[DC] Multiples - 'Same' proceedings scenarios for calculations testing](https://trello.com/c/ImE62qTS/95-dc-multiples-same-proceedings-scenarios-for-calculations-testing)

Davidslv:
This PR will be closed, only the example tests from separate convictions should be used.

Relevant orders has been covered in #438

PR #432, opened 2021-01-27
Refactor multiples calculator to work with proceedings

zheileman:
Instead of working with "check_group" objects, it is best to work with proceedings (meaning SameProceedings or SeparateProceedings) classes.

A bit of refactor to the calculator interface, so now instead of passing a check_group to the `#spent_date_for` method we pass the proceeding directly. Internally it knows what to do.

Also no need to call `#process!` manually anymore as that is now part of the initialization.

Finally, making sure the proceedings are returned always sorted by start date, in the "Check your answers" and in the results page.

For this sorting to work properly, I've reintroduce the simple caution "known date" step, otherwise we can't sort as for simple cautions we removed the date step.
The results page for simple caution however are left as before: `This caution is spent on the day you receive it`.

### DIFF
--- a/app/presenters/check_answers_presenter.rb
+++ b/app/presenters/check_answers_presenter.rb
@@ -3,16 +3,14 @@ class CheckAnswersPresenter
 
   def initialize(disclosure_report)
     @disclosure_report = disclosure_report
-
-    calculator.process! if disclosure_report.completed?
   end
 
   def summary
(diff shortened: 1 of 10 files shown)


Davidslv (approved):
👍

Davidslv:
Self note: this interferes with #430

PR #435, opened 2021-02-01
Add new conviction date step

zheileman:

We need to capture and extra date, which will be needed for proper multiple sentences inside a conviction calculations.

Up until now for these calculations we've been using the `known_date` but that is incorrect as that's the date the sentence started. We need the date of the conviction, which might be different.

This PR only adds the new DB fields and controller/view/form but it does not link it yet to the decision tree, that will be done in a follow-up PR.

Also, once we have all this wired up, we will need to update th multiples calculators (uh oh).

Screenshot 2021-02-01 at 12 15 09


Davidslv (approved):
👍

PR #436, opened 2021-02-01
Show conviction date step in multiples MVP

zheileman:

Ask this new date, but only if we are running the "multiples" MVP (so it will not show for simples or in production).
Revert some of the changes introduced in PR #429.

Unfortunately to do this we have to add some conditional code and propagate the `multiples_enabled` boolean to the decision trees as these do not have access to the controller/session and can't read the cookie.

In a follow-up PR the calculators will need to be updated.


Davidslv (approved):
👍

PR #438, opened 2021-02-02
Multiple convictions: Ignore relevant orders for same proceedings

Davidslv:
This commit modifies the calculations to ignore relevant orders.
This prevents a relevant order to drag-through in case it's the longest sentence in the group.
Removing the relevant order when it's not the longest won't affect the desired effect either as some other sentence will prevail.

It won't remove relevant orders when all sentences are relevant orders - it keeps the logic as it was before
It will keep the same logic as before when there are no relevant orders.

---

We need clarification to workout what is the correct behaviour when there are relevant orders,
Our current briefing on this matter is:

"if a conviction contains multiple sentences where one is a relevant order, the relevant order is not taken into account for drag-through purposes"

Although this does not clarify a scenario where there are 3 convictions and 2 convictions are relevant orders and are also the longest of the 3.

I've played the following scenario in Unlock calculator and they gave me the longest conviction (1st conviction)

They all started in 1/1/2000 (same proceedings)
1st has a length of 48 months (longet relevant order)
2nd has a length of 36 months (second relevant order)
3rd is a motoring fine, should have expired in 6 months without endorsement

I've created another card to revisit this calculations again, this card can be found here: https://trello.com/c/sZ7qBDwe/105-dc-double-check-that-the-calculations-for-multiple-relevant-orders-in-same-proceedings-are-correct

Story: https://trello.com/c/ImE62qTS/95-dc-multiples-same-proceedings-scenarios-for-calculations-testing

### DIFF
--- a/app/services/calculators/multiples/same_proceedings.rb
+++ b/app/services/calculators/multiples/same_proceedings.rb
@@ -18,42 +18,7 @@ module Calculators
       private
 
       def expiry_dates
-        expiration_dates
-      end
-
-      def expiration_dates
-        return expiry_date_for_all_sentences if all_are_relevant_orders? || no_relevant_orders?
-
-        expiry_date_without_relevant_orders
-      end
-
-      def all_are_relevant_orders?
-        disclosure_checks.all?(&:relevant_order?)
-      end
-
-      def no_relevant_orders?
-        disclosure_checks.none?(&:relevant_order?)
-      end
-
-      def relevant_orders
-        @_relevant_orders ||= disclosure_checks.select(&:relevant_order?)
-      end
-
-      def expiry_date_for_all_sentences
-        @_expiry_date_for_all_sentences ||= disclosure_checks.map(&method(:expiry_date_for))
-      end
-
-      def sentences_without_relevant_orders
-        (disclosure_checks - relevant_orders)
-      end
-
-      # When a conviction contains multiple sentences where one is a relevant order,
-      # the relevant order is not taken into account for drag-through purposes
-      # Given that same proceedings cares about the maximum date,
-      # we simply take away the relevant orders and return the expiry date of a non relevant order
-      # See the following card for more clarification, when provided: https://trello.com/c/sZ7qBDwe/
-      def expiry_date_without_relevant_orders
-        sentences_without_relevant_orders.map(&method(:expiry_date_for))
+        @_expiry_dates ||= disclosure_checks.map(&method(:expiry_date_for))
       end
     end
   end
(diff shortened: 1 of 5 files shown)


Davidslv:
[[DC] Multiples - 'Same' proceedings scenarios for calculations testing](https://trello.com/c/ImE62qTS/95-dc-multiples-same-proceedings-scenarios-for-calculations-testing)

app/services/calculators/multiples/same_proceedings.rb
```diff
@@ -18,9 +18,42 @@ def spent_date
...
 
       def expiry_dates
-        @_expiry_dates ||= disclosure_checks.map(
-          &method(:expiry_date_for)
-        )
+        expiration_dates
+      end
+
+      def expiration_dates
+        return expiry_date_for_all_sentences if all_are_relevant_orders? || no_relevant_orders?
```
zheileman, line 25: Could all this be simplified by basically obtaining a collection without the relevant orders, or, if empty, defaulting to all sentences, and then using the resulting dates to pick the max?

Kind of:

```ruby
def expiration_dates
  (sentences_without_relevant_orders.presence || disclosure_checks.presence).map(&method(:expiry_date_for))
end
```

zheileman (commented):
🥳 getting there!

I think we might be able to simplify it, take a look at the comment and see if it is possible, but other than that, and pending clarification on some of the edge cases, this LGTM.

Davidslv:
@zheileman commit 3d9a7cd is to prove (with test examples) that no code change were needed.

What I'm going to do is close this PR and open one specifically to add these scenarios to our tests so that we don't have more doubts about this.

I'm still to understand the connection in regards in how relevant orders in different convictions affect the calculations.

spec/factories/disclosure_check.rb
```diff
@@ -97,18 +97,23 @@
       conviction_with_known_date
       conviction_type { self.under_age.inquiry.yes? ? ConvictionType::YOUTH_MOTORING : ConvictionType::ADULT_MOTORING }
       conviction_subtype { self.under_age.inquiry.yes? ? ConvictionType::YOUTH_DISQUALIFICATION : ConvictionType::ADULT_DISQUALIFICATION }
+      conviction_length { 6 }
```
Davidslv, line 100: Will add this in a separate PR, they are needed

Davidslv:
[Ignore relevant order sentences in other separate, overlapping convictions.](https://trello.com/c/XJEnGNj6/109-ignore-relevant-order-sentences-in-other-separate-overlapping-convictions)

Davidslv:
> I'm still to understand the connection in regards in how relevant orders in different convictions affect the calculations.

This card has superseded this work and aims to explain the connection https://trello.com/c/XJEnGNj6/109-ignore-relevant-order-sentences-in-other-separate-overlapping-convictions

PR #439, opened 2021-02-02
Use the `conviction_date` in multiples calculator

zheileman:

Instead of the `start_date` (which is the `known_date` DB field), we need to use the new `conviction_date` because although some times both dates can be the same, when deciding if a spent date falls inside another rehabilitation period, we have to use the date of the conviction, not when the sentence or order started, as per legal clarification.


Davidslv (approved):
This looks good, thank you!

PR #441, opened 2021-02-02
Update factories

Davidslv:
I'm update the factories that will eventually be necessary to run more scenarios for the multiples calculations.

Extracted from #438


zheileman (approved):
👍

PR #443, opened 2021-02-03
More tweaks to the MultipleOffensesCalculator

zheileman:
Ticket: https://trello.com/c/PcAlcOaz

This work relates to the edge case where under certain circunstances a never spent, or indefinite conviction could take over the other convictions even if there was no overlapping between them.

In doing this I realised the loop can be simplified further, if we trick the special variants into behave like a Date in terms of ranges.

What this means is:
- A `ResultsVariant::NEVER_SPENT` is like an infinity date, it has no end.
- A `ResultsVariant::INDEFINITE` is a very far in the future date, but is not infinity (this difference is important as one takes precedence over the other).

There might be further tweaks/fixes to do and also it is still pending the work for relevant orders that _might_ affect the loop or the date we need to use on it. But that will come in separate PR(s).

### DIFF
--- a/app/value_objects/results_variant.rb
+++ b/app/value_objects/results_variant.rb
@@ -7,6 +7,17 @@ class ResultsVariant < ValueObject
     INDEFINITE   = new(:indefinite),
   ].freeze
 
+  # Needed so we are able to do ranges and compare dates with special variants
+  def to_date
+    case self
+    when NEVER_SPENT
+      Date::Infinity.new
+    when INDEFINITE
+      # absurdly in the future, but sooner than "infinity"
+      Date.new(2049, 1, 1)
+    end
+  end
+
   def self.values
     VALUES
   end
--- a/spec/value_objects/results_variant_spec.rb
+++ b/spec/value_objects/results_variant_spec.rb
@@ -12,4 +12,34 @@ RSpec.describe ResultsVariant do
       ))
     end
   end
+
+  describe '#to_date' do
+    subject { described_class.new(value) }
+
+    context 'when the variant is `never_spent`' do
+      let(:value) { 'never_spent' }
+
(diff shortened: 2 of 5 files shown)


app/value_objects/results_variant.rb
```diff
@@ -7,6 +7,17 @@ class ResultsVariant < ValueObject
     INDEFINITE   = new(:indefinite),
   ].freeze
 
+  # Needed so we are able to do ranges and compare dates with special variants
+  def to_date
+    case self
+    when NEVER_SPENT
+      Date::Infinity.new
+    when INDEFINITE
+      # absurdly in the future, but sooner than "infinity"
```
Davidslv, line 16: 🤣

spec/value_objects/results_variant_spec.rb
```diff
@@ -12,4 +12,24 @@
...
+      it 'considers an infinity date' do
+        expect(subject.to_date).to eq(Date::Infinity.new)
+      end
+    end
+
+    context 'when the variant is `indefinite`' do
+      let(:value) { 'indefinite' }
+
+      it 'considers a very far in the future date but not infinite' do
+        expect(subject.to_date).to eq(Date.new(2049, 1, 1))
```
Davidslv, line 31: I think we should add a test that fails if "Today" is 1 Jan 2049, so that someone can update the date to a even further future

zheileman: Yep good point, kind of a smoke test. Will add it.

app/value_objects/results_variant.rb
zheileman: Small reference to Blade Runner 😉😄

Davidslv (approved):

PR #445, opened 2021-02-04
Add multiple convictions test scenarios and tweak the calculator

zheileman:
Ticket: https://trello.com/c/PcAlcOaz

Note: this does NOT yet cover the relevant orders.

But in particular with this PR we are trying to cover these scenarios:

* If a conviction is received before another conviction is spent, neither conviction will become spent until the rehabilitation periods for both are over i.e. the longest one applies to both. To clarify some cases and exceptions:

  * If the first conviction has an end date and the second conviction is “never spent”, then the first conviction changes to “never spent”

  * If the first conviction is “never spent” and the second has an end date, then neither changes

The following scenarios are covered with tests and pass our current algorithm implemented in `MultipleOffensesCalculator`:

image (1)

---

Screenshot 2021-02-04 at 16 11 43

---

Screenshot 2021-02-04 at 16 12 45

### DIFF
--- a/app/services/calculators/multiples/multiple_offenses_calculator.rb
+++ b/app/services/calculators/multiples/multiple_offenses_calculator.rb
@@ -20,6 +20,7 @@ module Calculators
       def spent_date_for(proceeding)
         return unless disclosure_report.completed?
 
+        conviction_date = proceeding.conviction_date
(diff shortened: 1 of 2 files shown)


Davidslv (approved):
This is great!
Thank you 👏

PR #446, opened 2021-02-05
Ignore relevant orders on overlapping convictions 

Davidslv:
The solution proposed here is to have two different methods, the existing #spent_date
which is used to iterate through all sentences.

And the new method #spent_date_without_relevant_orders,
which means it will ignore any relevant order sentence and
output the spent_date of a non relevant order.

It's possible that there's only a relevant order left, so
in this case the #spent_date_without_relevant_orders will return nil.

Hence having to check for a nil spent_date in the algorithm of MultipleOffensesCalculator.

All doubles have been adapted to represent their existing scenarios.
More examples should be added to represent relevant orders.

Scenarios have been written based on examples supplied in:

- Nacro Guidance
- https://www.gov.uk/guidance/rehabilitation-periods
- https://hub.unlock.org.uk/knowledgebase/detailedguideroa

Story: https://trello.com/c/XJEnGNj6/109-ignore-relevant-order-sentences-in-other-separate-overlapping-convictions

### DIFF
--- a/app/models/disclosure_check.rb
+++ b/app/models/disclosure_check.rb
@@ -8,6 +8,9 @@ class DisclosureCheck < ApplicationRecord
   }
 
   def relevant_order?
+    # conviction subtype will be nil if it's a caution
+    return false if conviction_subtype.nil?
+
     ConvictionType.find_constant(conviction_subtype).relevant_order?
   end
 end
--- a/app/services/base_multiples_calculator.rb
+++ b/app/services/base_multiples_calculator.rb
@@ -33,11 +33,17 @@ class BaseMultiplesCalculator
   end
   # :nocov:
 
-  def without_relevant_orders
-    (disclosure_checks - relevant_orders)
+  def spent_date_without_relevant_orders
+    @_spent_date_without_relevant_orders ||= without_relevant_orders.map(
+      &method(:expiry_date_for)
+    ).max
   end
 
   private
+
+  def without_relevant_orders
+    (disclosure_checks - relevant_orders)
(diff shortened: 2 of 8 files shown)


app/services/calculators/multiples/same_proceedings.rb
```diff
@@ -15,6 +15,12 @@ def spent_date
         expiry_dates.max
       end
 
+      def spent_date_without_relevant_orders
```
zheileman, line 18: I think is worth moving this method to the `BaseMultiplesCalculator` and removing their individual implementations in the subclasses, because it will work for both, same proceedings, and separate proceedings:

- In same proceedings is what we have now so will continue working.
- In separate proceedings, when there is only 1 conviction, and it is a relevant order, the collection will be empty and a `max` will return `nil`.

zheileman (approved):
💯 🥳 Approved. It looks good, and I think we've covered all scenarios at least for now.

app/services/calculators/multiples/separate_proceedings.rb
```diff
@@ -5,6 +5,16 @@ def spent_date
         expiry_date_for(disclosure_check)
       end
 
+      def relevant_order?
```
Davidslv, line 8: this can go into a private method

Davidslv:
We are aware that there's another edge case and we agreed to fix it in a follow up PR.

The edge case is that it's considering the spent date of Conviction A as the spent date of Conviction C (as they overlap, but should not overlap with the relevant order)

image

PR #447, opened 2021-02-05
Refactor subclasses of BaseMultiplesCalculator

zheileman:
It is clear by now that having two separate classes, just to differentiate 1 sentence or more than 1 sentence is an overkill as the whole interface and methods are shared and can be made abstract to work in both cases.

Thus, I'm removing `SeparateProceedings`, `SameProceedings` and `BaseMultiplesCalculator` and organising all code in a single, generic class `Proceedings`.

Everything behaves the same, this is just an internal reorganisation.


Davidslv (approved):
Very nice and clean! 😎 👍

PR #448, opened 2021-02-05
Fixes overlapping convictions spent_dates

Davidslv:
We almost got it correct before.

To solve the issue of the scenario represented in the graph I had to compare the range of dates without the relevant order.

But first one needs to confirm if a conviction only holds a relevant order, if so, skip the range comparison.

If there's a date that does not belong to a relevant order, then we compare the overlapping, but never with relevant order dates.

My previous mistake was to only consider 1 proceeding, this now considers both proceedings

image

Story: https://trello.com/c/XJEnGNj6/109-ignore-relevant-order-sentences-in-other-separate-overlapping-convictions

### DIFF
--- a/app/services/calculators/multiples/multiple_offenses_calculator.rb
+++ b/app/services/calculators/multiples/multiple_offenses_calculator.rb
@@ -20,6 +20,10 @@ module Calculators
         # Cautions are always dealt with separately and do not have drag-through
         return spent_date unless proceeding.conviction?
 
+        # Comparison of dates between different convictions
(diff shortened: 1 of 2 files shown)


Davidslv:
[Ignore relevant order sentences in other separate, overlapping convictions.](https://trello.com/c/XJEnGNj6/109-ignore-relevant-order-sentences-in-other-separate-overlapping-convictions)

zheileman (approved):
🚢 💯 All green 😄

PR #449, opened 2021-02-08
Add approximate date copy to results page

zheileman:
Ticket: https://trello.com/c/L2pqJMHo

First part of the work, to add an additional line of copy when we detect there is at least one approximate date in the check.

The other line of copy, shown always, has been tweaked according to content designer.

In a follow-up PR the copy will be implemented in the hint texts.

Screenshot 2021-02-08 at 13 05 07


Davidslv (approved):
👍

PR #450, opened 2021-02-09
Amend hint text to clarify approximate dates

zheileman:
Ticket: https://trello.com/c/L2pqJMHo

Follow-up to PR #449.

We want to introduce some extra copy in all the dates hint text. However in doing this I realised we had many different hint texts depending if it was a caution/conviction and the type of the offence.

In order to make this more easy to maintain so we only have 1 hint text for all dates and it appears in the same place always, I've moved out from the hint text and into the `lead` paragraph some of the copy previously had in the hint text itself.

**Example 1**
Screenshot 2021-02-08 at 16 44 40

---

**Example 2**
Screenshot 2021-02-08 at 16 21 42

### DIFF
--- a/app/helpers/custom_form_helpers.rb
+++ b/app/helpers/custom_form_helpers.rb
@@ -25,6 +25,13 @@ module CustomFormHelpers
     ).html_safe
   end
 
+  # Used to customise lead text when reusing the same view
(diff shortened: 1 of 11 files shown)


Davidslv (approved):
👍

PR #451, opened 2021-02-09
Add more scenarios for multiple convictions

Davidslv:
These scenarios will be sent to legal to be checked, they have been numbered on purpose to be able to identify which test needs modifications - if that's the case.

We may add more so that we can understand which scenarios need attention, if that's the case do not change the order of the scenarios, just keep adding more without replacing existing ones.

This also removes all duplicate scenarios (1, 2 and 3) by deleting the file `written_scenarios_spec.rb`

Story: https://trello.com/c/6ynrKcLL/130-dc-multiple-conviction-scenarios


Davidslv:
[[DC] Multiple Conviction Scenarios](https://trello.com/c/6ynrKcLL/130-dc-multiple-conviction-scenarios)

zheileman (approved):
Amazing 🥳 💯

PR #452, opened 2021-02-10
Mechanism to share/return to the results page

zheileman:

This, for now, it is only for development/test purposes so it is hidden behind the "developer tools" footer menu.

Maybe in the future this is a feature we can expand and offer to end users so they can also share their results with other people.

The mechanism is simple. We use the ID of the `disclosure_report` to get the checks inside, and we pick one and assign their ID to the session, redirecting back to the results page to render the results by running the calculations again from the database.

As there is no personal identification at any point in the service, even if the link is shared publicly with unknown people, there is no risk at all, but just in case I'm using UUIDs so it is impossible to do enumeration attacks, for example.
Also once a "report" is completed, it cannot be altered/edited.

Reports are purged after 60 days so there is plenty of time to share, but we can also evaluate this in the future.

Screenshot 2021-02-10 at 11 09 33

### DIFF
--- a/app/controllers/concerns/error_handling.rb
+++ b/app/controllers/concerns/error_handling.rb
@@ -6,6 +6,8 @@ module ErrorHandling
       case exception
       when Errors::InvalidSession, ActionController::InvalidAuthenticityToken
         redirect_to invalid_session_errors_path
+      when Errors::ResultsNotFound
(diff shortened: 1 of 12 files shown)


Davidslv (approved):
This is awesome! 👍 🌮

PR #454, opened 2021-02-11
Improve multiples calculator with a memory (relevant orders fix)

zheileman:
Ticket: https://trello.com/c/AeZ9GHp0

This is supposed to "fix" yet another edge case, but using a very basic "memory" where we store the spent date after any drag-though so in the following iteration of the loop, we use it if present and keep dragging the date if needed (and store back again in the memory).

We've made the minimum amount of changes but there are other approaches.

image (2)

### DIFF
--- a/app/services/calculators/multiples/multiple_offenses_calculator.rb
+++ b/app/services/calculators/multiples/multiple_offenses_calculator.rb
@@ -11,6 +11,7 @@ module Calculators
         process!
       end
 
+      # rubocop:disable Metrics/AbcSize
       def spent_date_for(proceeding)
         return unless disclosure_report.completed?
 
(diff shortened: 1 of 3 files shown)


Davidslv:
[[DC] Add latest Simon's scenario](https://trello.com/c/AeZ9GHp0/131-dc-add-latest-simons-scenario)

Davidslv (approved):
💯

app/services/calculators/multiples/multiple_offenses_calculator.rb
```diff
@@ -36,7 +37,9 @@ def spent_date_for(proceeding)
         #
         convictions_by_date.without(proceeding).each do |conviction|
           other_conviction_date = conviction.conviction_date
-          other_spent_date = conviction.spent_date_without_relevant_orders
+
+          # Pick the spent date from memory if we have it, otherwise get it fresh.
+          other_spent_date = dates_memory.fetch(conviction, conviction.spent_date_without_relevant_orders)
```
Davidslv, line 42: Having a date in the memory means that the spent date of that conviction has been dragged through by another conviction. So instead of using the spent date of it's own sentences, it uses the date that has been stored in memory.

PR #458, opened 2021-02-15
Submit to sentry missing template exceptions

zheileman:
As part of ticket https://trello.com/c/HML7Aa42 we found out that Sentry by default was ignoring (not submitting) any `ActionView::MissingTemplate` errors so it was more difficult to find out where our code was failing.

Remove from the list of excluded exceptions `ActionView::MissingTemplate` so we can get these, including their backtrace.

### DIFF
--- a/config/initializers/raven.rb
+++ b/config/initializers/raven.rb
@@ -3,4 +3,9 @@ Raven.configure do |config|
   config.environments = %w( production )
   config.silence_ready = true
   config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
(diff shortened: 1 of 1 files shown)


Davidslv (approved):
👍

Davidslv:
[[DC][Bug] Error page is shown when user wants to add a new sentence/conviction and clicks back link](https://trello.com/c/HML7Aa42/137-dcbug-error-page-is-shown-when-user-wants-to-add-a-new-sentence-conviction-and-clicks-back-link)

PR #459, opened 2021-02-16
Remove error page when a check is completed

zheileman:
Ticket: https://trello.com/c/139idBDs

After some investigations it is quite possible this error happens when a user double click the continue button in the last of the questions thus first submit will mark the check as completed and second click will return error.

Removing this error will not have any impact internally (i.e. completed checks will still be marked as completed) but the user will not receive this error if they for example finish a check and go back with the browser and change some of their answers, no harm done.

We would let the user proceed and then if they've changed dates or something their results page will reflect these changes.

As this is a very self-contained commit/PR we could always revert if we find any issues in the future.


Davidslv (approved):
👍

PR #460, opened 2021-02-17
Add rehabilitation periods bar graphs

zheileman:
Ticket: https://trello.com/c/tClki61b

Add all the graphs that our (previous) service designer created for us, as they are quite visual and help understanding the different rehabilitation period rules and the drag-through.

More documentation that can help us (developers) can be place in the `/docs` directory.


zheileman:
@Davidslv If you can remember anything else that we could add to this please add it. I think I've added all the graphs but I might have missed some you may have?

Also what we can do next is adding a link in the spec scenarios to each of the corresponding graphs.

Davidslv (approved):
I think we have them all, I'll do the linking on the tests for you

PR #461, opened 2021-02-17
Revert removal of FPNs

zheileman:
This reverts commit ef87968bdfc518fd624b96e3a3de1748dad18d3c, reversing
changes made to bfa55a8d517e23a3ce8907053f1d0b000cfcaa7a.

Ticket: https://trello.com/c/Vcpqhlqr

As per policy and legal decision, instead of removing completely from
the service FPNs, we are going to maintain them but only when they are
given an endorsement and thus they become a conviction.

In order to do this as clean as possible, first I'm reverting the PR
where the FPNs were removed, fixing any conflicts, so it works as it
was before for `singles`.

It will not work for `multiples` as the work/refactors we did in the
multiples calculator was under the assumption we didn't have FPNs.

In a follow-up PR I will adapt the code so we only leave in the service
FPNs with endorsement, and this will work with multiples too.

### DIFF
--- a/app/decorators/conviction_decorator.rb
+++ b/app/decorators/conviction_decorator.rb
@@ -18,6 +18,11 @@ module ConvictionDecorator
       ConvictionType::ADULT_MOTORING.eql?(self)
   end
 
+  def motoring_penalty_notice?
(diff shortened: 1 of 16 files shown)


Davidslv (approved):
Thank you!

PR #463, opened 2021-02-19
Avoid jumping from h1 to h3 in results page

zheileman:
Ticket: https://trello.com/c/18AS69LI

Feedback from GDS:

Currently the first 2 headings appear to be h3s:
‘If a basic criminal record check is requested’
‘If a standard or enhanced criminal record check is requested’

These then skip to an h2:
‘How this result has been worked out’

Our internal accessibility guidance states:

'Each page should get one H1 (the title), followed by H2s and H3s. Don’t skip from an h1 to an h3. Nest headings appropriately so that screen reader users, who can navigate using a list of headings, don’t think they’ve missed one which doesn’t exist.'


Davidslv (approved):
👍 wasn't too bad, just a few of them

PR #464, opened 2021-02-24
Change back link to GOV.UK page

Davidslv:
### ⚠️ This is to be deployed tomorrow 1pm ⚠️ 

This is a requirement to be able to send users back to the
GOV.UK page that will be live tomorrow (25th February 2021, at 1pm).

The intention is to make the user feel it never left to a different platform.

Technically I decided to use a new class, to identify external URLs given that
we may have other external urls around, it would become easier to find and change them
in the future.

Story: https://trello.com/c/HPTKJ7Dp

### DIFF
--- a/app/views/steps/check/results/shared/_disclaimer.en.html.erb
+++ b/app/views/steps/check/results/shared/_disclaimer.en.html.erb
@@ -25,5 +25,5 @@
   start a new check.
 </p>
 
-<%= link_button :restart_check, root_path, class: 'govuk-button ga-pageLink govuk-!-margin-top-2',
+<%= link_button :restart_check, root_path(new: 'y'), class: 'govuk-button ga-pageLink govuk-!-margin-top-2',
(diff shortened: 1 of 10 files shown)


zheileman (approved):
👍 cool solution

app/views/steps/check/results/shared/_disclaimer.en.html.erb
```diff
@@ -25,5 +25,5 @@
   start a new check.
 </p>
 
-<%= link_button :restart_check, root_path, class: 'govuk-button ga-pageLink govuk-!-margin-top-2',
```
zheileman, line 28: There are many other `root_path` in use in the service, I think a better solution is to redefine `root` in the `routes.rb` to point to the new controller#action

That way there is no need to change one by one in all the views, etc.

Davidslv:
We decided to use redirection on the home page to keep the session warning and clear the session properly.
Back button points to Govuk page.

PR #467, opened 2021-03-11
Correction on Military Service Detention

Davidslv:
https://user-images.githubusercontent.com/136777/110816404-38c9ad80-8283-11eb-9580-cbf7a49a2460.mov

Currently we do not ask the length of a military service detention,
according to legal, we must ask for that length and then add 12 months for adults or 6 months for youth.

The correct interpretation is the following:

For adults, it's 12 months after the length given
For youth, it's 6 months after the length given

Story: https://trello.com/c/WzSPxKYw

### DIFF
--- a/app/value_objects/conviction_type.rb
+++ b/app/value_objects/conviction_type.rb
@@ -62,7 +62,7 @@ class ConvictionType < ValueObject
     DISMISSAL                          = new(:dismissal,                        parent: MILITARY, skip_length: true, calculator_class: Calculators::AdditionCalculator::StartPlusSixMonths),
     OVERSEAS_COMMUNITY_ORDER           = new(:overseas_community_order,         parent: MILITARY, calculator_class: Calculators::AdditionCalculator::PlusSixMonths),
     SERVICE_COMMUNITY_ORDER            = new(:service_community_order,          parent: MILITARY, calculator_class: Calculators::AdditionCalculator::PlusSixMonths),
-    SERVICE_DETENTION                  = new(:service_detention,                parent: MILITARY, skip_length: true, calculator_class: Calculators::AdditionCalculator::StartPlusSixMonths),
+    SERVICE_DETENTION                  = new(:service_detention,                parent: MILITARY, calculator_class: Calculators::AdditionCalculator::PlusSixMonths),
 
     REPARATION_ORDER                   = new(:reparation_order,                 parent: PREVENTION_REPARATION, relevant_order: true, skip_length: true, calculator_class: Calculators::ImmediatelyCalculator),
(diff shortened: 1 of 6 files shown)


Davidslv:
[[DC] Military convictions calculation error](https://trello.com/c/WzSPxKYw/192-dc-military-convictions-calculation-error)

app/value_objects/conviction_type.rb
```diff
@@ -62,7 +62,7 @@ def initialize(raw_value, params = {})
     DISMISSAL                          = new(:dismissal,                        parent: MILITARY, skip_length: true, calculator_class: Calculators::AdditionCalculator::StartPlusSixMonths),
     OVERSEAS_COMMUNITY_ORDER           = new(:overseas_community_order,         parent: MILITARY, calculator_class: Calculators::AdditionCalculator::PlusSixMonths),
     SERVICE_COMMUNITY_ORDER            = new(:service_community_order,          parent: MILITARY, calculator_class: Calculators::AdditionCalculator::PlusSixMonths),
-    SERVICE_DETENTION                  = new(:service_detention,                parent: MILITARY, skip_length: true, calculator_class: Calculators::AdditionCalculator::StartPlusSixMonths),
+    SERVICE_DETENTION                  = new(:service_detention,                parent: MILITARY, skip_length: false, calculator_class: Calculators::AdditionCalculator::PlusSixMonths),
```
zheileman, line 65: You can remove the `skip_length` key because the default is false. Only the exceptions will set it to `true`.

features/adults/conviction_military.feature
```diff
@@ -9,19 +9,20 @@ Feature: Conviction
     Then I should see "<known_date_header>"
 
     And I enter a valid date
-    Then I should see "<length_type_header>"
```
zheileman, line 12: I don't understand this. Why remove the variables?

zheileman (commented):
🎖️ 

I think the and length_type and conviction_length form headers need to talk about "detention" instead of "order"?
Similar to when we talk about "disqualification" or "discharge" etc.

features/adults/conviction_military.feature
Davidslv: I removed the variables because it felt that the repetition is on the table instead of the setup. 
The values are all the same, so I thought it would be less repetitive if we just add it straight into the scenario.

Davidslv: After updating the title, now it makes more sense to have the variables

features/adults/conviction_military.feature
```diff
@@ -9,7 +9,7 @@ Feature: Conviction
...
+    Then I should see "Was the length of the order given in weeks, months or years?"
```
zheileman, line 12: I think this also needs to be customised, `Was the length of the detention [...]` 🤣

zheileman (approved):
🥳

PR #470, opened 2021-03-18
Introduce drag through to relevant orders

Davidslv:
We have learned that not all relevant orders behave the same, meaning that some relevant orders do not drag through and others do.

The meaning of drag through is when 2 separate convictions share the same timeline. A relevant order that doesn't drag through ignores the overlap and ends when the conviction is spent. A relevant order that does drag through won't ignore the overlap and may be extended to the spent date of another conviction.

It's important to identify the relevant order, because relevant orders have an [additional length option](https://trello.com/c/MrEm9UKp).

The `drag_through` flag is to be used specifically for calculation purposes.

Story: https://trello.com/c/R9VWuOAu

### DIFF
--- a/app/models/disclosure_check.rb
+++ b/app/models/disclosure_check.rb
@@ -13,4 +13,11 @@ class DisclosureCheck < ApplicationRecord
 
     ConvictionType.find_constant(conviction_subtype).relevant_order?
   end
+
+  def drag_through?
+    # conviction subtype will be nil if it's a caution
+    return false if conviction_subtype.nil?
+
+    ConvictionType.find_constant(conviction_subtype).drag_through?
+  end
 end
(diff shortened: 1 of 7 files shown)


Davidslv:
[[DC] Conditional Discharge Orders (and any other Relevant Orders that are also subject to drag through)](https://trello.com/c/R9VWuOAu/146-dc-conditional-discharge-orders-and-any-other-relevant-orders-that-are-also-subject-to-drag-through)

Davidslv:
@zheileman I'm going to merge this so that Legal can go and test it on staging. Happy to address any concerns that you may see when you get a chance to review this PR

Davidslv:
> @zheileman I'm going to merge this so that Legal can go and test it on staging. Happy to address any concerns that you may see when you get a chance to review this PR

Can't really bypass this without a review 😄

njseeto (approved):
🚀

PR #471, opened 2021-03-23
Check Your Answers Page

Davidslv:
https://user-images.githubusercontent.com/136777/112176182-c8f6e380-8bef-11eb-925d-801ed0607e7a.mov

I've done a soft separation from CheckYourAnswers, which was being shared
between the results page and the Check Your Answers page.

Having MultipleResultsPresenter enable us to separate both pages, althought for now,
the only thing that is different is the `results_page` attribute that is passed
down the chain of existing classes.

I don't want to commit to fully separate both right now, it's not clear to me
at this moment what we will need once we start working on the results page, so
I rather have a soft transition here.

-----

I've also requested a decision to show number before the word "Conviction" (or "Caution"),
as this felt misleading to have the number after the word.

Prototype suggested

Conviction 1
Conviction 2

The problem emerged when a caution is added after a conviction, like so:

Conviction 1
Conviction 2
Caution 3

It could make the user think that they have a 3rd caution when they only inserted 1 caution.
So we have agreed to have the number before as if it was a numbered bullet point list.

1. Conviction
2. Conviction
3. Caution

-----

Header logic for sentence numeration

When a conviction has 1 sentence, do not show the header ("Sentence 1")
When a conviction has more than 1 sentence, show the header with numeration

Story: https://trello.com/c/eLQkrUGd

### DIFF
--- a/config/locales/en/helpers.yml
+++ b/config/locales/en/helpers.yml
@@ -173,6 +173,8 @@ en:
         kind_options:
           caution: You were given an official warning by the police
           conviction: You were found guilty of a crime in court
+      steps_check_caution_or_conviction_form:
+        add_caution_or_conviction_html: You must enter details of all cautious or convictions if you have received more than one.
       steps_caution_known_date_form:
         known_date_html: *date_hint_text
       steps_check_under_age_form:
@@ -251,6 +253,8 @@ en:
         under_age_options:
           'yes': Under 18
           'no': 18 or over
+      steps_check_caution_or_conviction_form:
+        add_caution_or_conviction_options: *YESNO
       steps_caution_known_date_form:
         approximate_known_date_options:
           'true': *approximate_date_checkbox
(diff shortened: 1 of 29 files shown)


Davidslv:
I'm adding a "do not merge" label until we have a direction on deployability.

Davidslv:
[[DC] Work on Check Your Answers Page](https://trello.com/c/eLQkrUGd/203-dc-work-on-check-your-answers-page)

zheileman (approved):
LGTM 👍 
As long as we remember to cleanup all the overhead, and the extra if-elses, and feature flag etc. I feel like we are adding more and more code just to protect what we have, when we should be scrapping most of it 🤣

config/locales/en/helpers.yml
```diff
@@ -307,6 +311,8 @@ en:
       steps_check_under_age_form:
         caution: How old were you when you got cautioned?
         conviction: How old were you when you got convicted?
+      steps_check_caution_or_conviction_form:
+        add_caution_or_conviction: Do you have another caution on conviction you'd like to check?
```
Davidslv, line 315: typo, usage of `on` between word `caution` and word `conviction`. Meant to be `or`

Davidslv:
Closing this now that we have #492

PR #472, opened 2021-03-24
Add a new caution or conviction Question Page

Davidslv:
**First and foremost**, I want to merge this PR into #471 instead of master, this causes a PR dependency but I'm ok with this as both should go together (but done separately) - also easier to review each individually too.

After check your answers page, we want to ask the user if they
have another caution or conviction before showing the results page.

Answering yes will start a new journey
Answering no will send the user to the results page

Story: https://trello.com/c/krAsstbg

https://user-images.githubusercontent.com/136777/112323061-55191180-8ca9-11eb-983c-9476368c83ba.mov

Add-on: 
Screenshot 2021-03-24 at 16 00 40


zheileman:
Have you tried to not select any radio option and click continue? It will blow up.
That's because this code does not follow the pattern, and you will be fighting against it 😿 

I recommend following the pattern where possible: controller -> form object -> update_and_advance -> decision tree.

If you don't want to create a DB attribute (I agree, for this it should not be necessary) you can override the `#save` method in the `CautionOrConvictionForm`.

zheileman (approved):
🥳

Davidslv:
Merging this into #471

PR #476, opened 2021-03-27
Implement Relevant Order edge case, Scenario 8 by Legal

Davidslv:
We have learned about another edge case around relevant orders, so far our understanding was that a relevant order would be ignored when comparing convictions.

Although, when a relevant order is the sentence with the longest period within it's own conviction AND is the longest when comparing with another conviction, it remains the longest in it's own conviction, and does not affect the external conviction.

For a better understanding, I've created and added a new graph, this can be seen in:
[docs/results/08_relevant_order_3.png](https://github.com/ministryofjustice/disclosure-checker/blob/ab9051deefa7f717e50b5b6a4b98314ea688de43/docs/results/08_relevant_order_3.png)

### Before

image

### After

image

### Video

https://user-images.githubusercontent.com/136777/112712242-0a281580-8ec6-11eb-8cc4-850bf5f5e78b.mov

Story: https://trello.com/c/ycbp7CgU


Davidslv:
[[DC] Scenario 8: Relevant Order (by Legal)](https://trello.com/c/ycbp7CgU/237-dc-scenario-8-relevant-order-by-legal)

zheileman (approved):

PR #473, opened 2021-03-25
🎏Feature flag enable multiples journey

Davidslv:
🎏This PR enables the multiples conviction journey. 🎏

We are getting this ready for next weeks deploy.
On hold until legal sign off

Will need copy updates from content designer, I will do those in a separate PR against this PR, then we can ship it all together, as I don't want to clog the deployment pipeline with this until we have sign off.

Story: https://trello.com/c/IimlbdWM

PR #478, opened 2021-04-07
Update GA transactions to cope with multiples

zheileman:
Ticket: https://trello.com/c/KF3PYWfT

Up until now we only had "single checks", as 1 check equals to just 1 caution or 1 conviction transaction on GA.

However going forward with multiples this needs to change, as 1 check (as in all the way to the results page) can equal to X number of convictions or cautions (added previously to a "basket").

As a first step so our current analytics don't start to get too skewed after the release of multiples we did in PR #473, we are adding a new transaction to the "CYA - Check your answer page" that triggers each time a new caution or conviction is added.

In addition, we modify the results page transaction so it contains information about how many proceedings it had and identify it by their report ID.

This is a first step and will be improved as we iterate and implement the new design and flow.

### DIFF
--- a/app/helpers/analytics_helper.rb
+++ b/app/helpers/analytics_helper.rb
@@ -1,6 +1,7 @@
 module AnalyticsHelper
   CUSTOM_DIMENSIONS_MAP = {
     spent: :dimension1,
+    proceedings: :dimension2,
   }.freeze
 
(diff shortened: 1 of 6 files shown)


njseeto (approved):
🚀

PR #480, opened 2021-04-09
Small cleanup and code reorganisation

zheileman:
Just some small cleanup of a functionality no longer in use (DISABLED_PARENT_TYPES) as well as code reorganisation.

Also removal of the hardcoded `drag_through: false` because the default is already false so no need to specify it, so it is easier to find the ones that are `drag_through: true` as these are just a handful.

Functionality is the same, this is only shuffling around some code.


njseeto (approved):
🚀

PR #482, opened 2021-04-16
Bump govuk_design_system_formbuilder gem

zheileman:
Updates the `govuk_design_system_formbuilder` gem to the latest version as the one we were using was a bit outdated.

There are several changes but nothing breaking. Our tests and cukes continue to pass and I did some manual testing as well with no obvious regressions.

Also starting with v2.2.0 the gem dropped support for Rails 5.2.4 so good thing we've upgraded to Rails 6.2 already.

Note: I had to lock `cucumber` to < 6.0.0 otherwise we would start getting an error, I suppose this is because some of the other gems are not yet updated to support cucumber > 6.0

We can test further on staging before releasing to production.


Davidslv (approved):
👍

PR #483, opened 2021-04-19
Do not track analytics for completed reports again

zheileman:
This is an improvement over a piece of work we did some time ago.

In essence, so we can use the "replay" functionality where we can see the results page at a later moment for a completed check, we need to avoid triggering some of the transactions and analytics otherwise it gets super skewed.

In order to do that, we add an extra attribute to the `disclosure_report` records, `completed_at` which is filled when the report is completed and then we can use this date later on to make decisions.

This will make it possible to replay old checks without tracking anything to GA.


Davidslv (approved):
👍

PR #484, opened 2021-04-20
Fix under reporting of proceedings in GA

zheileman:
We were reporting only cautions or convictions, but not the sentences/orders inside a conviction, which means a conviction with 3 sentences was reported just as 1 proceedings to google analytics, and in reality we want to know how many "multiples" checks are done and having more than 1 sentence qualify as "multiples".


Davidslv (approved):
Good spot 👍

PR #485, opened 2021-04-20
Rename adults community orders conviction group

zheileman:
Ticket: https://trello.com/c/7SK4PLNz

The work in the above ticket will be split into 2 parts. This PR is the first part.

Renaming of the `Community, prevention or reparation order` conviction group to `Community, reparation or other order with requirements`.

PR #486, opened 2021-04-20
Add new catch-all requirement order option

zheileman:
Ticket: https://trello.com/c/7SK4PLNz

This is the 2nd part of the work requested in the above ticket.

A new catch-all radio option in the "Community, prevention or reparation order" (note this bucket will be renamed once the PR #485 is merged) has been added, named "Any other order with a requirement".

At the moment this is a relevant order, with no drag through, and the calculation rules are the same as in the community order conviction.

However we need confirmation if these rules are correct before merging. Other than that the rest of the work is done.

Screenshot 2021-04-20 at 15 32 52

---

Screenshot 2021-04-20 at 15 33 09


Davidslv (approved):
👍

PR #488, opened 2021-04-21
Fix legal scenario 9

Davidslv:
This PR does two things:

1. Addresses the observation done by Legal that a Community Order shouldn't be a relevant order
> Youth rehab orders and Community orders are not relevant orders so are subject to the drag on effect. 

2. Fixes the scenario given by Legal by using the modified date (spent_date) as the comparison point between dates when this variable is modified from the initial assigned value.

> YRO is shown as spent on until 1 July 2016 but it should be showing as spent on 29 June 2021 along with the community order and custodial sentence.

This PR does not address the following quote:

> Also, it is not clear to me why the Referral order is prompting me to include the date that person attended first panel meeting. 
> My understanding is that referral orders like all other offences that do not have buffer periods run from date of conviction; and would be treated as running from date of conviction on basic checks. 

I'm not really sure what this means, but doesn't a breaking point right now (unlike point 1 and 2).

Further comment on commit e0ebe84.

Story: https://trello.com/c/r3BsvxQC

### DIFF
--- a/app/value_objects/conviction_type.rb
+++ b/app/value_objects/conviction_type.rb
@@ -78,7 +78,8 @@ class ConvictionType < ValueObject
     ######################
 
     ADULT_ATTENDANCE_CENTRE_ORDER       = new(:adult_attendance_centre_order,      parent: ADULT_COMMUNITY_REPARATION, calculator_class: Calculators::AdditionCalculator::PlusZeroMonths),
-    ADULT_COMMUNITY_ORDER               = new(:adult_community_order,              parent: ADULT_COMMUNITY_REPARATION, relevant_order: true, calculator_class: Calculators::AdditionCalculator::PlusTwelveMonths),
+    ADULT_COMMUNITY_ORDER               = new(:adult_community_order,              parent: ADULT_COMMUNITY_REPARATION, relevant_order: false, calculator_class: Calculators::AdditionCalculator::PlusTwelveMonths),
+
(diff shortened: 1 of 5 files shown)


app/value_objects/conviction_type.rb
```diff
@@ -78,7 +78,8 @@ def initialize(raw_value, params = {})
     ######################
 
     ADULT_ATTENDANCE_CENTRE_ORDER       = new(:adult_attendance_centre_order,      parent: ADULT_COMMUNITY_REPARATION, calculator_class: Calculators::AdditionCalculator::PlusZeroMonths),
-    ADULT_COMMUNITY_ORDER               = new(:adult_community_order,              parent: ADULT_COMMUNITY_REPARATION, relevant_order: true, calculator_class: Calculators::AdditionCalculator::PlusTwelveMonths),
+    ADULT_COMMUNITY_ORDER               = new(:adult_community_order,              parent: ADULT_COMMUNITY_REPARATION, relevant_order: false, calculator_class: Calculators::AdditionCalculator::PlusTwelveMonths),
```
Davidslv, line 81: I will remove the flag, as it's the default (relevant_order: false)

Davidslv:
[[DC] Investigate scenario given by legal](https://trello.com/c/r3BsvxQC/253-dc-investigate-scenario-given-by-legal)

zheileman:
giphy

zheileman (approved):
🤯 🥳 Great catch! let's hope is the last one 🤣

PR #489, opened 2021-04-22
Update Relevant Orders documentation

Davidslv:
This is just a mirror update of our [spreadsheet](https://docs.google.com/spreadsheets/d/1o7OligQtUKNIz_nzREb3PSUEQZje2by1_6BEKybGATc/edit#gid=583473806).

### DIFF
--- a/docs/relevant_orders.md
+++ b/docs/relevant_orders.md
@@ -2,31 +2,29 @@
 
 ## Adults
 
-| Category                                  | Conviction                     |
-| ----------------------------------------- | ------------------------------ |
-| Discharge                                 | Bind over                      |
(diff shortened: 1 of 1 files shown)


zheileman:
I grouped them by category and used the names in the service (for example Sexual harm prevention order) so it makes more sense, regardless of the spreadsheet 😢

zheileman (approved):

Davidslv:
> I grouped them by category and used the names in the service (for example Sexual harm prevention order) so it makes more sense, regardless of the spreadsheet 😢

Got it, I've updated to reflect the order and naming

PR #491, opened 2021-04-23
Big cleanup of the unused "singles" code, and small refactor

zheileman:
This is a big PR as otherwise nothing would pass CI and test would fail.

I've split it into a few commits to try to make more sense.

In essence this is the big cleanup of the "multiples" feature flag along with any related code, and simplification of existing controllers, presenters and views to just focus on multiples and the work we will need to do in coming weeks.

More simplification/refactor will be needed specially if we take some of the work done in PR #471 but this is a starting point.


Davidslv (approved):
This is looking good to me, great work 🎆 👍

PR #492, opened 2021-04-27
Refactor the CYA and Results presenters and views

zheileman:
Some big refactor and reorganisation of the different presenters we have.

Now there are separate presenters for CYA and for Results page, as well as different views. This will make it more flexible as we want to show different information/layout depending on which page we are.

Also simplification of some locales to avoid repetition.

Finally, the design of the CYA is now more similar to that of the User Research session (PR #471). I've salvaged what I could and refactored the rest.
The final design of the CYA is not yet 100% agreed but with this PR I hope we are closer to finish it, and also to be more easy to introduce changes.

I've not touched the results page, so it still is the old design (except for some minor typography changes).

Screenshot 2021-04-27 at 11 26 26

---

Screenshot 2021-04-27 at 11 26 57


Davidslv (approved):
This looks good to me, thank you 👍

PR #494, opened 2021-04-27
More design tweaks to CYA/results

zheileman:
Just adding the nice boxes around each of the proceedings (cautions or convictions) as I think this really makes a difference to clarify each of them.

Although the results page needs to change dramatically, I've added the box there too for consistency as it looks nicer.

This is following some of the latest changes to the prototype, so still WIP but starting to take form.

Screenshot 2021-04-27 at 15 45 02

--- 

Screenshot 2021-04-27 at 15 45 16

### DIFF
--- a/app/assets/stylesheets/local/custom.scss
+++ b/app/assets/stylesheets/local/custom.scss
@@ -10,6 +10,12 @@
   margin-bottom: 1em;
 }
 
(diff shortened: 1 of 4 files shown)


Davidslv (approved):
Really nice touch 👍

PR #466, opened 2021-03-04
Update journey to only have endorsed FPN

Davidslv:
Story: https://trello.com/c/jdhA6rjb/64-dc-endorsed-fpns-journey

We are not removing the page that questions users about endorsement, we are just removing non-endorsed FPN from the journey.

PR #495, opened 2021-04-28
Remove reference to non endorsed FPN

Davidslv:
Back in PR #466 we have agreed to only have endorsed FPN's in the 
service - this is due to non endorsed FPN not being considered a 
conviction. 

Now that we have the possibility to calculate multiple convictions, we 
no longer support non endorsed FPN's as they are not a conviction.

This is more of a follow up cleanup on #466


zheileman (approved):
💯

PR #496, opened 2021-04-28
Ensure we start a new session in some links

zheileman:
A very minor fix.

There were some edge cases where the links to start a new check would still show the message "you have a check in progress" which is not what we wanted.

This is because they were missing the "magic" parameter that indicates to not show that message and instead reset/start a new session.

Similarly, in the completed report error page, do not link to the results page if there are no completed checks as that makes no sense.

Renamed the helper method to make more clear what it does.


Davidslv (approved):
💯

PR #498, opened 2021-04-28
Add feature for multiple convictions

Davidslv:
This feature example covers a journey with two convictions (Discharge Order and Community Order).

I may be able to cut the number of steps in the future, right now I just wanted to have a feature test that goes around a second round.

### DIFF
--- a/features/multiples/cautions_and_convictions.feature
+++ b/features/multiples/cautions_and_convictions.feature
@@ -1,0 +1,72 @@
+Feature: A person with cautions and convictions
+
+  @date_travel
+  Scenario: Discharge Order and Community Order
+    Given The current date is 28-04-2021
+      And I have started a check
+      And I choose "Cautioned"
+      And I choose "Under 18"
+      And I choose "Youth caution"
+      And I enter the following date 01-01-2006
+     Then I should see "Check your answers"
+      And I click the "Add a caution or conviction" button
+
+      And I choose "Convicted"
+      And I choose "Under 18"
+      And I enter the following date 01-06-2008
+      And I choose "Referral or youth rehabilitation order (YRO)"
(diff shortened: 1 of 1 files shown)


zheileman (approved):
That's fantastic, precisely we are terribly lacking features about multiples so this is great!

As a suggestion, if you want to cut the number of steps and make this test more simple, perhaps you can get rid of almost all the `I should see "foobar"` because we cover all that in the many features we already have, so we can make some safe assumptions here.
After all, multiples is nothing more than "singles" with a different CYA/results page. But the journeys are all the same.

Another suggestion could be adding 1 young simple caution and 1 adult conviction (maybe with 2 sentences), that way you cover everything (youth and adults / cautions and convictions) at the same time, and also the end results page will give you 2 different dates so no need for the `twice` definition.
If you do this then the directory instead of adults/ could be just multiples/

But the important thing is to have something to start with 🥇

features/multiples/cautions_and_convictions.feature
```diff
@@ -0,0 +1,72 @@
+Feature: A person with cautions and convictions
+
+  @date_travel
+  Scenario: Discharge Order and Community Order
+    Given The current date is 28-04-2021
```
Davidslv, line 5: Had to add this purely only because when running the whole test suite there seems to be a Date leaking into this test which makes one of the results to be in the future.

> This conviction was spent on 10 December 2020

When running the whole test suite:

> This conviction will be spent on 10 December 2020

I will add a note to investigate where the leak is coming from

zheileman: Maybe in another feature we didn't properly tag the scenario with `@date_travel` and then after it is finished it didn't travel back to current time. Good catch 👍

PR #499, opened 2021-04-29
Refactor completion action before results page

zheileman:
Instead of performing the completion actions on the `#show` action of the results page, it is best to conform to a proper restful design and thus the "report" is marked as completed through a PUT request to the `reports_controller` (finish action).

After this, we redirect to the results page where there is no further logic or actions to perform, it just renders the view as expected, and can be reloaded as many times as the user wants.

It also makes the analytics "check completed" transaction a bit more explicit, as this transaction must only be triggered once. Before we had a workaround but was a bit convoluted, this is more solid.


Davidslv (approved):
👍

PR #501, opened 2021-05-06
Check we have a completed report before results page

zheileman:
There was a missing validation in the results page.

If a user, for whatever reason bookmarks the result page, next time they come back to the service they might not complete any check or leave midway a check and then jump straight to the results page.

It would show an empty page (as in no items: cautions or convictions in the basket) which would be weird.

With this validation we ensure we show the results page only if we have a "completed" report. If not we show an error message and let them either get back to the CYA page if they had something in the basket, or to start a new check.


Davidslv (approved):
nice!

PR #503, opened 2021-05-10
Results page new design

Davidslv:
Story: https://trello.com/c/QmITv0Rd

I've reconsiders the addition of the conviction date within the paragraph and found that it adds too much code for a minimal gain, because cautions do not have "conviction_date" but a "known_date". We already have that information on the table underneath so I think it's not worth the extra code for repeated information.

I'm also considering changing the name of "SpentDatePanel" but right now I don't have a better naming for it.

### Never spent
image

### Spent
image

### Unspent
image

### DIFF
--- a/app/presenters/spent_tag.rb
+++ b/app/presenters/spent_tag.rb
@@ -1,0 +1,34 @@
+class SpentTag
+  GREEN = 'green'.freeze
+  RED = 'red'.freeze
+
+  COLORS = {
+    ResultsVariant::SPENT => GREEN,
+    ResultsVariant::NOT_SPENT => RED,
+    ResultsVariant::NEVER_SPENT => RED,
+    ResultsVariant::SPENT_SIMPLE => GREEN,
+    ResultsVariant::INDEFINITE => RED,
+  }.freeze
+
+  attr_reader :spent_date
+
+  def initialize(spent_date:)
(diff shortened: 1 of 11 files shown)


Davidslv:
[[DC] Front-end work for cosmetic changes on Results page](https://trello.com/c/QmITv0Rd/217-dc-front-end-work-for-cosmetic-changes-on-results-page)

app/presenters/spent_tag.rb
```diff
@@ -0,0 +1,38 @@
...
+  }.freeze
+
+  attr_reader :spent_date
+
+  def initialize(spent_date:)
+    @spent_date = spent_date
+  end
+
+  def color
+    if spent_date.instance_of?(Date)
```
zheileman, line 20: Is it possible here you just do `COLORS[variant]` as `#variant` method already do their magic?

zheileman (approved):
👍 🥳 
Looking good! Only thing I can think of is to check in mobile/responsive view, just in case.

Davidslv:
Seems all ok on mobile (also used Chrome browser)

image
image

app/presenters/spent_tag.rb
Davidslv: Good shout! 👍

PR #504, opened 2021-05-11
Refactor and DRY some code in preparation for DBS

zheileman:

Note: There will be a follow-up PR, didn't want this to grow too much with the introduction of the DBS code.

We have exposed an `#spent?` method in the `ResultsVariant` so we can simplify some code.

Also, instead of repeating the `variant` or `tense` code snippet to know if a given date is in the past or not, we just pass to the `SpentTag` and to the (future) `DbsVisibility` classes the variant, as these don't need to know the specific `spent_date` or do any calculation with it.


Davidslv (approved):
👍

PR #505, opened 2021-05-12
Show DBS information on results page (feature-flagged)

zheileman:

Follow-up to PR #504.

Implement the DBS informational inset block on the results page, but for now it will only show on local/staging envs until we have the green light.

Also, the rules for enhanced checks are not completely clear, so my best attempt was when in doubt, to say `may appear`.

Basic checks are correct to the best of my knowledge. Also enhanced youth cautions.

Screenshot 2021-05-12 at 10 24 25

### DIFF
--- a/app/presenters/check_group_presenter.rb
+++ b/app/presenters/check_group_presenter.rb
@@ -28,6 +28,14 @@ class CheckGroupPresenter
     )
   end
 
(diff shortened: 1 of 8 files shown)


Davidslv (approved):
👍 👍

PR #508, opened 2021-05-13
Copy amendments and introduction of convictions logic

Davidslv:
Update from Leigh:

1. The line for approximate dates should appear under the opening line beginning: “Your results are calculated using…”
2. The line for approximate dates should read: As you entered an approximate date, your results will be approximate.
3. This paragraph is conditional to be shown only when it’s a conviction: [CONDITIONAL FOR CONVICTIONS] Your results are correct if you served your sentence in full as the court ordered you to. The conviction might not be spent if you did not stick to the terms of your sentence.

Story: https://trello.com/c/ONHg9IzX/278-dc-results-page-missing-copy

### Screenshot with a conviction

image

### Screenshot with a caution

image

### DIFF
--- a/app/presenters/results_presenter.rb
+++ b/app/presenters/results_presenter.rb
@@ -11,6 +11,10 @@ class ResultsPresenter < BasketPresenter
     ConvictionType::ADULT_MOTORING
   ].map(&:to_s).freeze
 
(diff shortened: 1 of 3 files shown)


zheileman (approved):
👍

PR #509, opened 2021-05-13
Cleanup and refactor `ValueObjectMethods` module

zheileman:
A bit of cleanup and refactor of some unnecessary complexity in the `ValueObjectMethods` (we were abusing it, including the module en many places).

Now, it is possible to do `disclosure_check.caution` or `disclosure_check.conviction` to get the corresponding Caution or Conviction.

This can simplify a lot the code but in order to not make this PR too large I've only changed some obvious places, there are many others that we can change.

Also removed an obsolete `#time_on_bail?` method.


Davidslv (approved):
This is really good 👍 Thank you!

PR #510, opened 2021-05-14
Add DBS rule for custodial convictions

zheileman:

Add the rule for custodial convictions (these are always shown on the enhanced DBS check).

Simplified/refactored some more code to make use of a more declarative syntax, easier to read, introduce in previous PRs.


Davidslv (approved):
👍

PR #512, opened 2021-05-18
Update results page survey

zheileman:
Ticket: https://trello.com/c/PslLguBi

Update the end of service survey with a new one as provided by user researcher.

As this survey is longer (it has an extra question) I've had to tweak the styling of it so it shows in full length.


Davidslv (approved):
👍

PR #515, opened 2021-05-19
Bump GOV.UK Frontend

zheileman:
Ticket: https://trello.com/c/gr75HUCY

Version 3.12.0 of the GOV.UK Frontend was released recently.

https://github.com/alphagov/govuk-frontend/releases/tag/v3.12.0

Main change is links are now easier to read and have a clearer hover state. This is opt-in (in future releases will be enabled by default).


Davidslv (approved):
👍

PR #517, opened 2021-05-21
Minor copy improvements for multiple checks

zheileman:
Ticket: https://trello.com/c/55y9BdCo

A couple of minor copy addition/tweaks to better signify the fact users can check more than 1 caution or conviction.

Copy as per content designer.

Ideally to be released ASAP before any further undergoing CYA work.


Davidslv (approved):
👍

PR #518, opened 2021-05-24
Functionality to remove sentences in CYA

Davidslv:
This adds functionality for the user to be able to remove sentences in the Check Your Answers page.

When a sentence is deleted, it's crucial to know wether there was other sentences
in the same report. Specially when the deleted sentence was the last one to be created,
since, until now, the session relied purely on the last `disclosure_check`, if that is deleted
we must find the Report id in order to know if there was other sentences present in order to send
the user to CYA or to the first page of the service (and restart their journey).

image

Story: https://trello.com/c/D6Goc7YV

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,8 +11,14 @@ class ApplicationController < ActionController::Base
     payload[:user_agent] = request&.user_agent
   end
 
+  # With the ability to remove disclosure_check records we now look for
+  # a disclosure_report when we can't find the last created disclosure_check.
+  # This guarantees that any other disclosure_check records added by the user,
+  # will be available to that user.
   def current_disclosure_check
-    @_current_disclosure_check ||= DisclosureCheck.find_by_id(session[:disclosure_check_id])
+    @_current_disclosure_check ||=
+      DisclosureCheck.find_by_id(session[:disclosure_check_id]) ||
+      DisclosureReport.find_by_id(session[:disclosure_report_id])&.disclosure_checks&.last
   end
   helper_method :current_disclosure_check
 
@@ -40,6 +46,7 @@ class ApplicationController < ActionController::Base
   def initialize_disclosure_check(attributes = {})
     DisclosureCheck.create(attributes).tap do |disclosure_check|
       session[:disclosure_check_id] = disclosure_check.id
+      session[:disclosure_report_id] = disclosure_check.disclosure_report.id
     end
   end
 end
--- a/app/forms/steps/check/remove_check_form.rb
+++ b/app/forms/steps/check/remove_check_form.rb
@@ -1,0 +1,24 @@
+module Steps
+  module Check
+    class RemoveCheckForm < BaseForm
+      include SingleQuestionForm
+
+      validates_presence_of :remove_check
+
+      yes_no_attribute :remove_check
+
+      private
+
+      # We do not store this in the database, so when user answers "No" we return true
+      # Otherwise if the record is successfully destroyed, return true
+      def persist!
+        raise DisclosureCheckNotFound unless disclosure_check
+
+        return true if remove_check.no?
+
+        disclosure_check.destroy
+        disclosure_check.destroyed?
+      end
+    end
+  end
+end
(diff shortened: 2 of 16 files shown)


app/services/check_decision_tree.rb
```diff
@@ -22,4 +24,13 @@ def after_under_age
       edit('/steps/conviction/conviction_date')
     end
   end
+
+  def after_remove_check
+    if disclosure_check.disclosure_report.disclosure_checks.any?
```
zheileman, line 29: Probably best to change to `disclosure_check.disclosure_report.disclosure_checks.completed.any?` (note the _completed_ scope) to ensure we only return to the CYA page if there are any other completed cautions or convictions otherwise in the case of kick out pages etc. there would be left overs and the user will go back to an empty CYA.

app/controllers/application_controller.rb
```diff
@@ -11,8 +11,14 @@ def append_info_to_payload(payload)
     payload[:user_agent] = request&.user_agent
   end
 
+  # With the ability to remove disclosure_check records we now look for
+  # a disclosure_report when we can't find the last created disclosure_check.
+  # This guarantees that any other disclosure_check records added by the user,
+  # will be available to that user.
   def current_disclosure_check
-    @_current_disclosure_check ||= DisclosureCheck.find_by_id(session[:disclosure_check_id])
```
zheileman, line 15: I'm feeling like there might be a simpler solution for this 😞 
The fact we need to look for one or the other, and set 2 IDs in the session feels itchy.

For now is ok but hopefully we can come up with a better solution.

app/forms/steps/check/remove_check_form.rb
```diff
@@ -0,0 +1,22 @@
+module Steps
+  module Check
+    class RemoveCheckForm < BaseForm
+      include SingleQuestionForm
+
+      yes_no_attribute :remove_check
```
zheileman, line 6: Validation error (when not selecting any of the radios) is not set in the locales.

zheileman (changes requested):
Looking good! 🥳 
A few comments and also we might need to give a think to see if there is a better way to do the session handling.

app/controllers/application_controller.rb
Davidslv: I'm hoping to rewrite it later, I think for now it's a nice middle step, then can focus on having DisclosureReport to be the starting point (and session) - I think!

zheileman: Cool, let's leave this for now. In the work I'm doing for the "change" functionality I'm also facing a similar problem where I need to refer to different disclosure_check record ID depending of the "change" link being clicked, so maybe there is some common ground between the 2 (remove and edit) and we can improve this.

app/views/steps/check/check_your_answers/shared/_check_row.html.erb
```diff
@@ -1,6 +1,15 @@
...
-</h3>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-one-half">
+    <h3 class="govuk-heading-m govuk-!-margin-bottom-2">
+      <%=t "#{check_row.kind}.orders.#{check_row.order_type}", scope: check_row.scope %>
+    </h3>
+  </div>
+
+  <div class="govuk-grid-column-one-half app-text-align-right">
+    <%= link_to('Remove this sentence', edit_steps_check_path(check_row.disclosure_check), class: 'govuk-link govuk-link--no-visited-state app-link-red govuk-!-font-size-19') %>
```
zheileman, line 9: Upps, just realised, for cautions this might need to read "Remove this caution" and for convictions not everything is a sentence so again we have the problem of naming things 🤣 Wonder if "Remove this check" is good enough 😞

zheileman (approved):

app/views/steps/check/check_your_answers/shared/_check_row.html.erb
Davidslv: maybe "Remove this one" / "Remove this"

Davidslv:
image

Davidslv:
[[DC] CYA with remove sentence functionality](https://trello.com/c/D6Goc7YV/294-dc-cya-with-remove-sentence-functionality)

PR #519, opened 2021-05-25
Check your answers `change` functionality

zheileman:

First attempt and kind of Proof of Concept of the ability to `change` answers to previous questions in cautions or convictions.

Right now this PR only implements Cautions as these are much simpler (only 2 questions can be edited).

Some questions will not be editable because in doing so, the caution (or conviction) change fundamentally, for example the under age question. If a user answered over 18 and then they change it to under 18, the whole caution or conviction will become wrong as the "type" is linked to the age (for example, youth cautions vs. adult cautions).

The implementation is simple: all of the "change" links will have as a query param the ID of the `disclosure_check` record being edited (because when in the basket there are more than 1, we need to differentiate).
This ID allows to "swap" the session from whatever was before to this new disclosure check record and perform the edit/navigation normally on it.

There is no possibility of tampering with the ID because it does the lookup over the records owned by the current disclosure report anyways, just in case.

More refinements, cukes, and convictions, in follow-up PRs...

Screenshot 2021-05-25 at 10 54 25

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -35,6 +35,13 @@ class ApplicationController < ActionController::Base
 
   private
 
+  def swap_disclosure_check_session(other_check_id)
+    session[:disclosure_check_id] = current_disclosure_report.disclosure_checks.find_by!(id: other_check_id).id
+
+    # ensure we don't have a memoized record anymore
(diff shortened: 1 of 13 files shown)


Davidslv (approved):
👍 nice!

PR #520, opened 2021-05-25
Add validation to caution date

zheileman:
For conditional cautions, the "known_date" (caution date) must be before the conditions end date.

This so far has been fine but with the `change` functionality in the CYA, the user might edit a conditional caution date and introduce a wrong date without noticing.

We might need to revisit other places where this can happen in convictions for the same reason, but for now I'm focusing on cautions as per PR #519.

### DIFF
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -9,6 +9,7 @@ en:
               blank: Enter the date of the caution in the format dd/mm/yyyy
               invalid: The date is too far in the past. Enter a date after 01/01/1940
               future: The date of the caution can’t be in the future
+              before_conditional_date: The date of the caution can’t be after the date conditions ended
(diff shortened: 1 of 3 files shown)


config/locales/en/errors.yml
```diff
@@ -9,6 +9,7 @@ en:
               blank: Enter the date of the caution in the format dd/mm/yyyy
               invalid: The date is too far in the past. Enter a date after 01/01/1940
               future: The date of the caution can’t be in the future
+              before_conditional_date: The date of the caution can’t be after the date conditions ended
```
Davidslv, line 12: I'm not too sure on this text

Davidslv (approved):
all good, just not sure on the text presented, but can be tweaked later on

PR #521, opened 2021-05-26
Some attributes can jump back to CYA

zheileman:
Some attributes can be edited without having to go through the following steps. This way after changing something, we can take the user back to the CYA page quickly.

This is done by marking these attributes with the `?next_step=cya` query param.

After editing something and clicking the `continue` button, if the `next_step` was set, the decision tree logic will be skipped and jump straight to the path specified in the next step (it is a hidden tag in the form).

If the attribute is the last one in a journey, there is no need to add this query param as it is redundant because the next step is always the CYA in these cases.
For example for cautions, `conditional_end_date` is the last attribute in the journey and the next step is always the CYA anyways.

### DIFF
--- a/app/helpers/custom_form_helpers.rb
+++ b/app/helpers/custom_form_helpers.rb
@@ -1,4 +1,6 @@
 module CustomFormHelpers
+  delegate :params, :hidden_field_tag, to: :template
+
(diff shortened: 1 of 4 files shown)


Davidslv (approved):
👍

PR #522, opened 2021-05-26
Delete disclosure report session

Davidslv:
Introduced in PR #518, I forgot to  delete the `disclosure_report` session so that the user can start a new check. Without  deleting the session the user would be prevented from starting a new check without deleting the service session cookies manually.


zheileman (approved):
🥳

PR #523, opened 2021-05-26
Change conviction attributes in CYA

zheileman:

Follow-up to PR #519.

This PR adds the "change" functionality to the convictions, similar to cautions.

There is one without the change link for now (`conviction_date`) because it is the top level conviction date, so changing it would affect all the disposals/orders that are part of the conviction (we would have to iterate through all of them changing the attribute in their corresponding DB records).

Will investigate some alternatives in another PR if possible. But it is best probably to not allow this to be changed.

Screenshot 2021-05-26 at 12 50 24


Davidslv (approved):
👍

PR #524, opened 2021-05-26
Fix color contrast in CYA remove link

zheileman:
Before:
Screenshot 2021-05-26 at 14 41 13

After:
Screenshot 2021-05-26 at 15 11 03


Davidslv (approved):
Makes sense 👍

PR #525, opened 2021-05-26
When removing a record, change the session if possible.

Davidslv:
Upon removing a disclosure check record, if there's other records within the same report, we swap them so that the user continues to iterate through other records.

This deprecates the second session that was introduced in 84ce9d77038c860dc0a6650e69b9db5eea7ed4f5 and uses the swap functionality introduced in 252db539d8261e8f3cab9d9bc7f822aff3df3574.


zheileman (approved):
👍 Looks more clear to me now! Thanks!

PR #526, opened 2021-05-27
A few more markups improvements

zheileman:
Convert the proceeding heading (caution / conviction name) to a definition list so the remove link move automatically down below the name when on small screens.

Also simplify a bit the styles of the red remove link.

Desktop:
Screenshot 2021-05-27 at 15 34 10

Mobile:
Screenshot 2021-05-27 at 15 33 26


Davidslv (approved):
nice 👍

PR #527, opened 2021-06-01
Remove endorsement question for motoring disqualifications

zheileman:
Ticket: https://trello.com/c/1RtNkemH

During DC demos, feedback was given that it is unnecessary to ask users who enter that they have a driving disqualification whether they have an endorsement, as all disqualifications are endorsed.

Legal confirmed that we should remove the question for the service.

Refactor the code around the disqualification logic because we now only have 1 motoring conviction where we need to still ask this question (fines). The others can be simplified.

### DIFF
--- a/app/decorators/conviction_decorator.rb
+++ b/app/decorators/conviction_decorator.rb
@@ -26,14 +26,11 @@ module ConvictionDecorator
     ].include?(self)
   end
 
-  def motoring_penalty_notice?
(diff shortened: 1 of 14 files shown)


Davidslv (approved):
👍 Looks good to me.

For a moment I thought we had forgotten the rule on the Youth Penalty Points (3 years) over the endorsement (2.5years) but gladly it was well commented

PR #529, opened 2021-06-07
Minor copy amendments to results page

zheileman:
As instructed by content designers.
There was a duplicated sentence and a wrong link.


Davidslv (approved):
👍

PR #530, opened 2021-06-10
Add cucumber scenario for CYA change links

zheileman:
Ticket: https://trello.com/c/ABI0WCOR

Add a simple cucumber scenario to cover the change links functionality on the check your answers page, similar to what we did with the remove functionality.

Note how we can nicely write these by using, for example, `I click the "Change start date" link` despite on the page only the word `Change` is visible, the `start date` part is visually hidden, but still it can be used to target elements with cucumber.


Davidslv (approved):
Perfect 👍

PR #533, opened 2021-06-24
Remove compensation to a victim DBS receipt kick out

zheileman:
Ticket: https://trello.com/c/cETYcVq3

This is work in progress. We are awaiting for copy changes (maybe) to be
introduced in the journey.

For now I've just reorganised the journey to get rid of the kick out
page and the amount question, as it was only used to trigger the
kick out when the amount was over 100 but does not affect any of the
calculations.

Pending things: removal of form objects, controllers, views, locales,
etc. and even the unused DB fields. once we are clear on the copy
changes and where we have to apply them.


Davidslv (approved):
👍 thanks

PR #534, opened 2021-06-24
Cleanup compensation receipt kick out leftovers

zheileman:
Follow-up to PR #533 to remove the controllers, form objects, views, etc. of the now removed receipt kick out.

In a separate PR there will be some more cleaning, including DB attributes.


Davidslv (approved):
👍

PR #535, opened 2021-06-24
Cleanup compensation paid amount leftovers

zheileman:
Similar to PR #534, remove the code that is no longer needed, together with the DB attributes.

This is the last PR for this work.


Davidslv (approved):
thanks, nice cleanup 🧼

PR #537, opened 2021-06-24
Introduce days to sentence length

Davidslv:
We have received feedback that sometimes courts give the sentence in days,
which means that users would have to convert those days into one of our options,
this is not ideal as it could prone users to introduce decimal numbers (which are not allowed).

So this commit allows users to introduce days, and internally we convert those days into months.

I've removed the conversion constants and replaced the calculations with ActiveSupport own methods.

The reason the specs have been changeds is that ActiveSupport calculates seconds per month,
so there's was a tiny discrepancy with the previous implementation, but nothing drastic (at least not in short periods of time)

```ruby
ActiveSupport::Duration::SECONDS_PER_MONTH

1.month.in_seconds
=> 2629746
```

## Question Page

image

## Results Page

image

Story: https://trello.com/c/DqoWLnHu

### DIFF
--- a/app/forms/steps/conviction/conviction_length_form.rb
+++ b/app/forms/steps/conviction/conviction_length_form.rb
@@ -4,7 +4,9 @@ module Steps
       attribute :conviction_length, String
       delegate :conviction_length_type, to: :disclosure_check
 
-      validates_numericality_of :conviction_length, greater_than: 0, only_integer: true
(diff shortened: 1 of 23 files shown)


Davidslv:
[[DC] Accepting days for sentence length](https://trello.com/c/DqoWLnHu/334-dc-accepting-days-for-sentence-length)

zheileman (approved):
Fantastic! 🥳

PR #538, opened 2021-06-25
Bump design system version

zheileman:
Bump govuk-frontend to latest version 3.13.0.

It includes bug fixes. No breaking changes.

In particular it stops the thick hover underlines from skipping the descenders, which was very unpleasant visually.


Davidslv (approved):
👍

PR #539, opened 2021-06-25
Improve handling of session on CYA page

zheileman:
Ticket: https://trello.com/c/gtvCFFpA

There was a nasty edge case where the session could become invalid (as in, pointing to a non-existing DB record) if, in the CYA, a user pressed any of the buttons to add a new sentence or a new caution/conviction but didn't complete this new journey, leaving an incomplete check (record) in the DB.

When returning to the CYA and pressing the `continue to results` button, then there is code to purge any incomplete checks from DB, purging also the record currently in the session and thus failing to render the results page (with the generic error `session expired`).

Fixed this by making sure after the purge of any incomplete checks, we re-assign an existing check back to the session from those in the report (the last one, but could be any, doesn't matter).

Also, as an improvement to this, if the user in the CYA press button to add a new caution/conviction, they will start a new journey but the `back` link would have take them to the govuk start page which is not the expected behaviour.
Now this back link will take them back to the CYA as expected.

There is something else to improve: when completing a few convictions, and in the CYA, clicking the header will not give the warning as usual and will wipe the session and start from scratch, which could be very annoying to somebody having entered many convictions already. Will be fixed/improved as a separate PR as is a bit unrelated to this one.

### DIFF
--- a/app/controllers/concerns/completion_step.rb
+++ b/app/controllers/concerns/completion_step.rb
@@ -26,6 +26,12 @@ module CompletionStep
   # remove any incomplete checks as they don't serve any purpose now
   def purge_incomplete_checks
     current_disclosure_report.disclosure_checks.in_progress.destroy_all
+
+    # It could happen the current session was one of the purged
+    # incomplete checks, so this ensures we have a valid one again.
+    swap_disclosure_check_session(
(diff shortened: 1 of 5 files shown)


Davidslv (approved):
Nice one! Thank you 🥇

PR #541, opened 2021-06-29
Show check in progress warning when on CYA

zheileman:
Ticket: https://trello.com/c/gtvCFFpA

Follow-up to PR #539 to fix/improve another issue, where the CYA didn't have the "check in progress" protection, so if a user had added let's say a few cautions/convictions in the basket and for some reason they click the header, it will just start a journey from the beginning, wiping from the session all previous entered cautions/convictions, which can be very frustrating.

This has been improved by not only checking for a "disclosure_check" record `in_progress` but also a disclosure_report, because when the user reach the CYA, the current disclosure_check is marked as completed and thus this is why the in progress warning wasn't showing.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -8,8 +8,8 @@ class HomeController < ApplicationController
   private
 
   def in_progress_enough?
-    current_disclosure_check&.in_progress? &&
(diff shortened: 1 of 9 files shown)


Davidslv (approved):
👍 good spot, thank you!

PR #542, opened 2021-06-30
Cleanup old participants functionality

zheileman:
In the past, when we opened up the service to a limited set of users as a trial, we had a mechanism for granting these users (normally YOTs) a unique set of credentials to access the service and we also tracked how many of them accessed and used the service.

This is not needed anymore since the service is now public for all without credentials.

Big cleanup of the remaining code, and prepare for repurpose the back office for new functionality (TBD.)


Davidslv (approved):
Nice cleanup!

PR #543, opened 2021-06-30
Add a checks report to the back office

zheileman:
Very simple, and using what we had already in place, report page in the back office, listing the last 50 completed reports, ordered by completion date, with some basic information and a link to show the results.

This is behind http auth credentials (same credentials as usual) and also is totally anonymous as we don't ask or collect any personal information.

It is just a starting point in case we want to extend/improve it. And is just a page we can remove if we don't use it or need it anymore.

Screenshot 2021-06-30 at 15 16 13


Davidslv (approved):
👍 thank you

PR #545, opened 2021-07-02
Ability to sort the report by number of sentences

zheileman:
Just a small improvement so it is now possible to sort the completed reports by its number of sentences

This way is easier to find at the top those that are more interesting because have many convictions/sentences.

Added a new `sentences` column to the report table, because 1 conviction, can have more than 1 sentence. It also counts as sentence the cautions, although this is probably not technically correct, but for the purpose of the report it makes more sense.


Davidslv (approved):
💯

PR #547, opened 2021-07-26
Implement DBS recency rules

zheileman:
Ticket: https://trello.com/c/lK5NwwEM

In addition to the rules we already have for the enhanced DBS checks, we are now introducing the recency ones too, as follows:

```
- Adult cautions given within 6 years of the check: will appear.
- Youth convictions imposed within 5.5 years of the check: will appear.
- Adult convictions imposed within 11 years of the check: will appear.
```

This only affects some **spent caution/convictions**.

We keep saying `maybe` if none of the rules are triggered, as a precaution.

**Note**: pending copy updates in the results page to indicate the results are valid as of the day of the check. Will be done in a separate PR when we have confirmation.

### DIFF
--- a/app/decorators/conviction_decorator.rb
+++ b/app/decorators/conviction_decorator.rb
@@ -16,6 +16,14 @@ module ConvictionDecorator
     ].include?(parent)
   end
 
+  def youth?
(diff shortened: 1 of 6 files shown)


Davidslv (approved):
👍 looks good

PR #548, opened 2021-07-27
Tweak DBS recency rules for older adult cautions

zheileman:
Follow-up to PR #547.

We are still awaiting confirmation but it is quite possible spent adult cautions given more than 6 years ago will not show in the standard or enhanced DBS check.

If given within 6 years they will appear.

More context in the slack convo: https://mojdt.slack.com/archives/CKZT0MY1G/p1627375025048900

As a consequence of this I had to do some refactor of the code but now it should be even more easier to read and follow and also easier to make changes in the future if needed.

### DIFF
--- a/app/presenters/dbs_visibility.rb
+++ b/app/presenters/dbs_visibility.rb
@@ -20,23 +20,20 @@ class DbsVisibility
     :will
   end
 
-  # Enhanced check rules:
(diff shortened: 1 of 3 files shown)


Davidslv (approved):
👍

zheileman:
This has been signed off, merging to master.

PR #549, opened 2021-07-27
Copy changes in results page

zheileman:
This work is related to the DBS recency rules introduced.

As part of this our content designer has identified a few changes to be done and some other copy to be refined in the results page, particularly to mention the fact the `results are valid on the date of the check`.

Screenshot 2021-07-27 at 13 19 18


Davidslv (approved):
👍

PR #550, opened 2021-08-24
Minor copy changes

zheileman:
Ticket: https://trello.com/c/WpAWw9Sc

A few copy changes suggested by our content designer.

There are a couple copy changes remaining that needs discussion, will be made in a separate PR when decided.

PR #553, opened 2021-09-08
Explain DBS `may appear` in results page

zheileman:
Ticket: https://trello.com/c/WpAWw9Sc

This is a follow-up to PR #550. There was a missing paragraph that has been implemented in this PR.

However this PR might not be required, just leaving here until we get confirmation from content designer.


Davidslv (approved):
👍

PR #567, opened 2022-01-21
Avoid content repetition in results page

zheileman:
Ticket: https://trello.com/c/qRnvt0Oy

The results page might repeat a whole chunk of copy if there were more than 1 caution or conviction that `may appear` on enhanced DBS.

Using `flush` fixes the problem and will only show the content once, even if `content_for` was called more than once.

### DIFF
--- a/app/views/steps/check/results/shared/_dbs_visibility.html.erb
+++ b/app/views/steps/check/results/shared/_dbs_visibility.html.erb
@@ -3,7 +3,7 @@
   enhanced_result = dbs_visibility.enhanced
 
   content_for(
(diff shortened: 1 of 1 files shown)


Davidslv:
For future reference, when the `flush` option is true, `content_for` replaces the previous content instead of concatenating it. An example is given [in here](https://api.rubyonrails.org/v7.0.1/classes/ActionView/Helpers/CaptureHelper.html#method-i-content_for) (search for flush).

Davidslv (approved):
👍 good one!

PR #570, opened 2022-01-28
Modify link anchor in results page

zheileman:
Ticket: https://trello.com/c/8NJm6gNh

On the back of the above ticket, it was decided a quick and immediate action we can take is to modify the DBS links to go to the paragraph "Using your results" as otherwise users, specially on mobile devices might miss it and contains important information about the DBS.

Previously links would jump to the underneath section, skipping the "Using your results" paragraph that may contain important information depending on the DBS results.


Davidslv (approved):
👍

PR #571, opened 2022-02-01
Protect against some uninitialized objects

zheileman:
Ticket: https://trello.com/c/i6m6IrA2

This is part of the investigation to fix or mitigate some (very small in number) exceptions that are captured in Sentry.

These errors happen when a user (could be even a bot scanning the website) access a page after the session has ended and thus there was no protection against a potential `nil` object.

I might raise a separate PR for other kind of errors.


Davidslv (approved):
👍

PR #592, opened 2022-11-25
CDPT-393 Validate current disclosure check is valid for step

vertism:
An exception can occur when a user reaches an invalid step for the current check. e.g. they reach a conviction step when it should be checking a caution. They are also not able to resume the previous check because the navigation stack is invalid.

To recreate:
1. Select "Cautioned" as the check type
2. Change the URL to `/steps/conviction/conviction_subtype`

The additional validation added means that if this occurs, the user will be redirected to the reset_session page which will allow them to start a new check, or resume the previous one. The invalid step is removed from the navigation stack so that they can continue.


Polsonby (approved):
Nice work!

PR #593, opened 2022-12-02
Maintenance page

vertism:
Adds a maintenance page that will show when the database is unavailable. This will be the case when DB upgrades are being performed.

When various exceptions occur, the user is redirected to an error page. However, this redirect logic is also applied when the user is already on an error page which can lead to an infinite redirect loop. There is now a check so that redirects will not occur if the user is already on an error page.


Polsonby (approved):
LGTM :)

PR #607, opened 2023-03-31
Disable sentry running when not in production like environment

vertism:
Don't send exceptions to sentry in test and local environments


njpreddy (approved):
Looks good to me.

PR #643, opened 2023-06-13
CDPT-742 Use rubocop-govuk

vertism:
Uses base config with one addition to exclude `lib/generators` as these are written in an unusual way that rubocop struggles with.

Exceptions to the rules have been used sparingly, but are needed in a few places where significant changes would be required to comply.


lucas-shaw (approved):
LGTM

PR #670, opened 2023-09-13
Update rehabilitation periods

vertism:
Update rehabilitation periods in calculators to reflect new legislation.
Also refactored so that each subclass has its own `rehabilitation_period` method.


njpreddy (approved):
Awesome, LGTM

PR #672, opened 2023-09-14
Deploy main after test run

vertism:
Uses `workflow_call` so deploy to Staging when PR is merged to main only occurs after tests have been run.


MeenaModhvadia (approved):
👍

PR #673, opened 2023-09-14
Fix workflow deadlock

vertism:
The value of `github.workflow` was the same for both workflow when one is called via `workflow_call` so this caused a deadlock and stopped the deploy working.

Hardcoding the names to be different should fix this.


MeenaModhvadia (approved):
👍

PR #675, opened 2023-09-15
Fix rehab 12 month bug

vertism:
Dates over exactly 12 months and under 13 months were not getting handled by the case statement.


njpreddy (approved):
LGTM

PR #676, opened 2023-09-18
CDPT-935 Schedule 18 calculation

vertism:
Adds new calculators for prison and detention sentences that are for schedule 18 offences.

Also refactors SuspendedPrison calculator to inherit from Prison, and DetentionTraining to inherit from Detention.


lucas-shaw (approved):
LGTM

PR #678, opened 2023-09-19
CDPT-967 Fix calculation bug

vertism:
Sentence lengths over 6 months and below 7 months are not being caught correctly by the case statement.

NB: I have not added any tests for this fix as this code is changing anyway as part of the upcoming CDPT-954 changes.


njpreddy (approved):
LGTM

PR #679, opened 2023-09-19
CDPT-958 Check answers change

vertism:
Removes separate calculators for Schedule 18 offences.

New `schedule18_over_4_years` method on disclosure_check determines a "yes" or "no" result based on the answers given. If none of the questions are answered then the result is blank and the question is not show on the check answers page.


lucas-shaw (approved):
LGTM

PR #689, opened 2023-09-26
Screen one content change schedule18

lucas-shaw:
As per the Jira ticket https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?modal=detail&selectedIssue=CDPT-982, additional wording required for the Schedule 18 policy source link

### DIFF
--- a/app/forms/steps/conviction/conviction_schedule18_form.rb
+++ b/app/forms/steps/conviction/conviction_schedule18_form.rb
@@ -4,6 +4,12 @@ module Steps
       include SingleQuestionForm
 
       yes_no_attribute :conviction_schedule18
+
+      # As we reuse this form object in multiple views, this is the attribute
+      # that will be used to choose the locales for legends and hints.
+      def i18n_attribute
+        conviction_subtype
+      end
     end
   end
 end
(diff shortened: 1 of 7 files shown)


app/forms/steps/conviction/conviction_schedule18_form.rb
```diff
@@ -4,6 +4,12 @@ class ConvictionSchedule18Form < BaseForm
       include SingleQuestionForm
 
       yes_no_attribute :conviction_schedule18
+
+      # As we reuse this form object in multiple views, this is the attribute
+      # that will be used to choose the locales for legends and hints.
+      def i18n_attribute
```
vertism, line 10: Good to use the existing conventions.

As the text on the page doesn't vary depending on the subtype, maybe it would make sense to use a static value here, and then only one translation value is needed?

Also, a test should probably be added for this method.

config/locales/en/helpers.yml
```diff
@@ -407,3 +409,4 @@ en:
         adult_suspended_prison_sentence: *order_started_lead_text
         adult_disqualification: ''
         youth_disqualification: ''
+
```
vertism, line 412: This file is missing the newline at the end of the file. 
You should be able to setup your editor to automatically add it if it's missing when you save a file.
It can also be setup to remove any unnecessary whitespace.

vertism (approved):
LGTM

PR #693, opened 2023-09-29
Cdpt 939 screen three

MeenaModhvadia:


app/services/conviction_decision_tree.rb
```diff
@@ -35,8 +37,14 @@ def destination
 
 private
 
+  def after_single_sentence_length
```
vertism, line 40: I don't think you need this method if it always takes you to check_your_answers.

There aren't any tests for it which is why you are getting a coverage error, so removing it will fix that too!

vertism (approved):
LGTM!

PR #694, opened 2023-10-04
CDPT-1015 Add/Remove Drag Through

vertism:
Changes are based on legal advice from Government Legal Department
Also renames variable to `no_drag_through` as a true value means that drag through does not apply.


lucas-shaw (approved):
LGTM

PR #697, opened 2023-10-16
CDPT-1049 Fix drag through calculations

vertism:
Convictions that were overlapped by earlier convictions were not having drag through correctly applied.


lucas-shaw (approved):
LGTM

PR #698, opened 2023-10-16
CDPT-1059 Enable maintenance mode

vertism:
A maintenance page will appear in place of every page on the site.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -53,6 +53,6 @@ private
   end
 
   def maintenance_mode
-    render "errors/maintenance"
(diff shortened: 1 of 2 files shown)


wilson1000-MoJ (commented):
Hi @vertism , I'm wondering: should we direct the User to a different service on Gov.UK?

The following page is direct to our service; maybe we should explain why their journey has been stopped.
https://www.gov.uk/tell-employer-or-college-about-criminal-record/check-your-conviction-caution

wilson1000-MoJ (approved):
Looks good 👍🏽

PR #706, opened 2023-10-30
CDPT-1111 Remove content and link to add more convictions

vertism:


app/views/steps/check/check_your_answers/_add_more.html.erb
wilson1000-MoJ, line 1: Good morning 👋🏽 
Is this intended to be empty?

wilson1000-MoJ (approved):

PR #708, opened 2023-10-30
CDPT-1115 Change military commmunity order rehabilitation periods

vertism:
Changes to match civilian community orders:
- Adds "until further order" option
- rehabilitation period matches end date given by the order


wilson1000-MoJ (approved):
Looks good 😄 great work, Andrew.

PR #728, opened 2024-02-22
CDPT-1385 Update crown

vertism:
Updates the crown in the logo by adding govuk-components and upgrading govuk-frontend.
Doing this also required updating govuk_design_system_formbuilder.

There were some small changes required to support these upgrades, and it also means the fonttype has changed a little bit, as has the look of the phase banner.


lucas-shaw (approved):
LGTM

PR #729, opened 2024-03-01
Fix error page when DB is down

vertism:
Allow error page to work when dev_tools_enabled? is enabled and the database is not working.


MeenaModhvadia (approved):
👍

PR #747, opened 2024-04-23
Add permissions for each deploy step

vertism:
Add permission, and pull image before re-tagging.
Also uses per environment variables that are now been set from cloud platform and updates replica numbers per environment.


njpreddy (approved):
LGTM

PR #771, opened 2024-08-05
Update Dockerfile & README

vertism:
Also removes unused gems, set start script in deployment terraform, adds specific platforms to lockfile, changes how removes unnecessary env vars.


njpreddy (approved):
LGTM

PR #797, opened 2024-09-10
Cdpt 1859 dc drag through changes

MeenaModhvadia:
Multiple convictions can have a drag through flag set to ensure convictions are only spent when sentence date expired for all convictions.


spec/factories/disclosure_check.rb
```diff
@@ -198,5 +198,14 @@
     trait :in_progress do
       status { :in_progress }
     end
+
+    trait :with_service_reprimand do
```
vertism, line 202: What is this used for?

spec/value_objects/conviction_type_spec.rb
```diff
@@ -495,6 +505,15 @@
       it { expect(conviction_type.calculator_class).to eq(Calculators::AdditionCalculator::PlusZeroMonths) }
     end
 
+    context "when ADULT_MILITARY" do
```
vertism, line 508: Should this be "ADULT_REPRIMAND"?

vertism (approved):
LGTM

PR #802, opened 2024-09-25
CDPT-2058 DC: Penalty points no drag through and rehabilitation period

MeenaModhvadia:
Penalty points set to no drag through
Rehabilitation period for penalty points amend to 3 years.


vertism (approved):
LGTM

PR #803, opened 2024-09-25
CDPT-2056 DC - Driving Disqualification Rehabilitation period changes

MeenaModhvadia:
Driving disqualification rehabilitation to be amended to the period of disqualification.


spec/services/calculators/disqualification_calculator_spec.rb
```diff
@@ -19,26 +19,20 @@
     let(:under_age) { GenericYesNo::YES }
 
     describe "#expiry_date" do
-      context "with a lengthless than or equal to 2.5 years" do
+      context "with a length less than or equal to 2.5 years" do
```
vertism, line 22: These contexts don't seem relevant anymore as the calculation doesn't change based on under or over a certain length.

spec/services/calculators/disqualification_calculator_spec.rb
```diff
@@ -19,26 +19,20 @@
...
 
-      context "with a lengthgreater than 2.5 years" do
+      context "with a length greater than 2.5 years" do
         let(:conviction_length) { 3 }
         let(:conviction_length_type) { "years" }
 
         it { expect(calculator.expiry_date.to_s).to eq((known_date + 36.months).to_s) }
       end
 
-      context "without a length" do
```
vertism, line 36: Has there been a change for convictions without a length?

MeenaModhvadia: A disqualification with no length is spent on day of conviction.
Shall I put a test for that?

vertism: I guess if that is really a thing that exists, then there should be a test for it.

MeenaModhvadia: I have tried to resolve this with a test but my changes now give the result as 'no length' instead of spent for motoring disqualification on day of conviction.
 Hospital order also tested with no length and result given 'no length'  which may be correct I need to check.
 
 I think this is wrong

spec/services/calculators/disqualification_calculator_spec.rb
vertism, line 1: There is no test for when the disqualification has a length.

app/services/calculators/disqualification_calculator.rb
vertism, line 1: There doesn't appear to be any reason to have separate Adult and Youth classes anymore.

In fact, should the Calculators::AdditionCalculator::PlusZeroMonths be used instead?

app/value_objects/conviction_type.rb
```diff
@@ -73,7 +73,7 @@ def initialize(raw_value, params = {})
 
     YOUTH_DISQUALIFICATION             = new(:youth_disqualification,         parent: YOUTH_MOTORING, relevant_order: true, no_drag_through: true, calculator_class: Calculators::DisqualificationCalculator::Youths),
     YOUTH_MOTORING_FINE                = new(:youth_motoring_fine,            parent: YOUTH_MOTORING, skip_length: true, calculator_class: Calculators::Motoring::Youth::Fine),
-    YOUTH_PENALTY_NOTICE               = new(:youth_penalty_notice,           parent: YOUTH_MOTORING, skip_length: true, calculator_class: Calculators::Motoring::Youth::PenaltyNotice),
+    YOUTH_PENALTY_NOTICE               = new(:youth_penalty_notice,           parent: YOUTH_MOTORING, skip_length: true, no_drag_through: false, calculator_class: Calculators::Motoring::Youth::PenaltyNotice),
```
vertism, line 76: Should this be changed? The ticket doesn't mention penalty notices.

app/services/calculators/addition_calculator.rb
```diff
@@ -1,16 +1,29 @@
...
       def added_time
         {}
       end
     end
 
+    #
+    # If disqualification and no length is given: conviction start date
+    #
+
+    class PlusZeroMonthsDisqualification < AdditionCalculator
```
vertism, line 17: I would suggest calling this StartPlusZeroMonths to match the other classes

app/value_objects/conviction_type.rb
```diff
@@ -103,9 +103,9 @@ def initialize(raw_value, params = {})
...
 
-    ADULT_DISQUALIFICATION               = new(:adult_disqualification,               parent: ADULT_MOTORING, relevant_order: true, no_drag_through: true, calculator_class: Calculators::DisqualificationCalculator::Adults),
+    ADULT_DISQUALIFICATION               = new(:adult_disqualification,               parent: ADULT_MOTORING, relevant_order: true, no_drag_through: true, calculator_class: Calculators::AdditionCalculator::PlusZeroMonthsDisqualification),
     ADULT_MOTORING_FINE                  = new(:adult_motoring_fine,                  parent: ADULT_MOTORING, skip_length: true, calculator_class: Calculators::Motoring::Adult::Fine),
-    ADULT_PENALTY_NOTICE                 = new(:adult_penalty_notice,                 parent: ADULT_MOTORING, skip_length: true, calculator_class: Calculators::Motoring::Adult::PenaltyNotice),
+    ADULT_PENALTY_NOTICE                 = new(:adult_penalty_notice,                 parent: ADULT_MOTORING, skip_length: true, no_drag_through: false, calculator_class: Calculators::Motoring::Adult::PenaltyNotice),
```
vertism, line 108: Why has this changed?

app/services/calculators/addition_calculator.rb
vertism: It also needs to be tested

vertism: Adding tests will help you figure out why this isn't working yet.

app/services/calculators/disqualification_calculator.rb
```diff
@@ -1,57 +1,23 @@
...
-      return ResultsVariant::INDEFINITE if indefinite_length?
-
-      if disclosure_check.conviction_length?
-        conviction_start_date.advance(rehabilitation_with_length)
-      else
-        conviction_start_date.advance(rehabilitation_without_length)
+    #
+    # If disqualification and no length is given: conviction start date is spent date
+    #
+    class StartPlusZeroMonths < AdditionCalculator
```
vertism, line 6: What is the purpose of this class and why is it inheriting from AdditionCalculator?

app/services/calculators/addition_calculator.rb
```diff
@@ -1,9 +1,8 @@
 module Calculators
   class AdditionCalculator < BaseCalculator
     ADDED_TIME_IF_NO_LENGTH = { months: 24 }.freeze
-
+    #
     # If length is given: conviction end date
-    # If no length is given: conviction start date + 24 months
```
vertism, line 6: Did you mean to remove this comment?

spec/services/calculators/disqualification_calculator_spec.rb
```diff
@@ -15,67 +14,22 @@
...
-      end
-
-      context "with an indefinite length" do
-        let(:conviction_length_type) { "indefinite" }
-
-        it { expect(calculator.expiry_date).to eq(ResultsVariant::INDEFINITE) }
-      end
+  describe "#expiry_date" do
+    context "without a length" do
+      it { expect(calculator.expiry_date.to_s).to eq("2018-10-31") }
```
vertism, line 19: If you look at the old tests they were set up so that the expectation is "known_data + some.duration". I think that is clearer to show what the expectation is rather than just a date with no context.

vertism (approved):
LGTM

PR #859, opened 2025-02-18
CDPT-2458 Fix prison and suspended sentence rehabilitation dates

njpreddy:
**Description**

  Prison sentence & Suspended Prison sentence rehabilitation date end is incorrect. 
  For both these sentences an extra day needs to be added so for example instead of ending on 31st will end on the 1st.

**Jira ticket**

  https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?selectedIssue=CDPT-2458


vertism (approved):
LGTM

PR #861, opened 2025-03-04
Re-add notification banner

vertism:
The banner had been disabled.
It has also been updated to match what is on production


HettieS (approved):
I couldn't find the ticket to cast my eye over the content, but LGTM

